PR #19, opened 2015-09-28
Update ContentResponseResult.java

Denix80:
Missed method


eugeniosant:
Hi Denix. Maybe it's better to use an implementation like this 
https://github.com/entando/entando-core/blob/master/engine/src/main/java/org/entando/entando/aps/system/services/guifragment/api/GuiFragmentResponse.java
It's more compliant with other classes.

nicpuddu:
closed because we never got a reply back from the user

PR #21, opened 2015-09-28
Corrected search_result widget in case of no search parameter

ahew:
I corrected the **search_result** widget GUI by placing a check in case the **search** parameter is null


eugeniosant:
Hi!! Maybe the line 409 would be better if you replace it with the following code:
<#if (RequestParameters.search?? && RequestParameters.search!='''')>
...or...
<#if (RequestParameters.search??)>

ahew:
<#if (RequestParameters.search?? && RequestParameters.search!='''')>

eugeniosant:
Thanks, Good Job.
For your next pull request, it's better to create a topic branch for the commits you want to propose.
https://help.github.com/articles/using-pull-requests/

PR #32, opened 2016-02-12
Fixed problem in concurrent usage of singleton ApsEntityDOM

ahew:
Added a method clone in ApsEntityDOM, invoked just before building the DOM.
The method instantiate the current class (ex: ContentDOM or FormDOM) and copy only the rootElementName so, if necessary, you only need to extend the method copyInto(IApsEntityDOM).


eugeniosant:
Thanks Elia

PR #37, opened 2016-04-20
fix issue 1 template label into menu.jsp

mbolzan:
Into the admin-console menu.jsp there isn't the translation s:text of "Template"


eugeniosant:
Thanks Marco

PR #46, opened 2016-11-21
Fixed generation of an url by always encoding in UTF-8 the parameters

ahew:
These changes allow to avoid the injection, in an url, of parameters having forbidden characters, both in names and in values.

The method
`public String createUrl(IPage requiredPage, Lang requiredLang, Map<String, String> params)`

is changed. Before it called:
`return this.createUrl(requiredPage, requiredLang, params, true, null);`
hiding the escape of & in parameters, now, it calls:
`return this.createUrl(requiredPage, requiredLang, params, false, null);`

This change have no effects because this method is called whitout parameters or with only one parameter.


entando-service:
All is well

entando-service:
All is well

PR #49, opened 2016-12-11
Evo i18n params

ahew:
Improvement for the i18n system:
 * Rendering method in the manager, instead of into the tag
 * Ability to replace parameters dynamically in manager, tag and content models

In the **I18nManager**, I used the class

`org.apache.commons.lang3.text.StrSubstitutor`

to replace parameters having format **${<PARAM_NAME>}**

**Example**

For example, I have the label **LABEL_WITH_PARAMS** with:

- en label: **Label with parameters ${surname} - ${name}**

- it label: **Etichetta con parametri ${name} - ${surname}**

----

**I18n Tag**

In the I18nTag, I used, nested, the wp:parameter tag, so I can write something like:

```
<wp:i18n key="LABEL_WITH_PARAMS">
	<wp:parameter name="name" >My Name</wp:parameter>
	<wp:parameter name="surname" >My Surname</wp:parameter>
</wp:i18n>
```
having:

- en result: **Label with parameters My Surname - My Name**

- it result: **Etichetta con parametri My Name - My Surname**

----

**Content Model**

I improved the content model with a method

`getLabelWithParams(<LABEL_CODE>)`

that returns a

`I18nLabelBuilder`

object having a method 

`I18nLabelBuilder addParam(<PARAM_KEY>, <PARAM_VALUE>)`

that returns the same object (enabling more than one parameter substitution).  
The final replace is done in the **toString()** method

Following the previous example, I can write:

```
getLabelWithParams("LABEL_WITH_PARAMS").addParam("name", "My Name").addParam("surname", "My Surname")
```

having the same result has using the i18n tag


eugeniosant:
Hi Elia. Good Job.
Can you resolve the conflicting files?

entando-service:
All is well

entando-service:
All is well

eugeniosant:
Good Job Elia!

PR #86, opened 2017-05-11
add better management for page_preview_hash parameter

stefanopuddu:


engine/src/main/java/org/entando/entando/aps/system/services/page/PageTokenManager.java
```diff
@@ -108,15 +110,6 @@ public String decrypt(String property) {
...
 	}
 
-	public void updateHash() throws Exception {
-		try {
-			String param = this.generateRandomHash();
-			this.generateInternalSaltAndPassword(param);
-		} catch (Throwable t) {
-			throw new Exception("error in updateHash", t);
-		}
-	}
```
nicpuddu: did we check if we had any other class actually calling this public method before deleting it?

stefanopuddu: yes we did. this property cannot be changed by design (at least for now)

PR #89, opened 2017-05-11
 label added ( they may conflict with other labels) on page designer + structure updated

DMinnai:


admin-console/src/main/java/com/agiletec/apsadmin/portal/package_it.properties
```diff
@@ -59,7 +59,7 @@ name.pageCode=Codice
 name.pageTitle=Titolo
 name.pageModel=Modello
 name.hasDefaultWidgets=Applica un set predefinito di <span lang="en">Widget</span>
-name.isShowablePage=Mostra questa Pagina nei men\u00f9
+name.isShowablePage=Mostra questa Pagina nei menu'
```
nicpuddu: lets revert back to entities please.

PR #109, opened 2017-05-17
Update hypertext attribute link: url resource

mcarta-rt:
Update hypertext attribute link: url resource


entando-service:

PR #121, opened 2017-05-23
ENQT-196: Add check permission on dashboard functionality

eugeniosant:


nicpuddu:
this looks fine but I don't believe it completely fixes the issue with the add button in all the three areas of the UI, does it?

eugeniosant:
You're right. I added the missing controls

nicpuddu:
:shipit:

PR #156, opened 2017-06-06
fix breadcrumbs menu and never touched jsps

DMinnai:


entando-service:
All is well

entando-service:
All is well

PR #157, opened 2017-06-06
fix configure page action with virtual root

stefanopuddu:
fix configure page action when a virtual root is present.


entando-service:
All is well

entando-service:
All is well

PR #164, opened 2017-06-06
bug fix, treeBuilder check root node

mcarta-rt:
bug fix, treeBuilder check root node


entando-service:
All is well

entando-service:
All is well

PR #166, opened 2017-06-07
ENQT-350 - Update property tag

eugeniosant:
Update property tag of struts2 core; change default value for attribute escapeHtml (from true to false)


entando-service:
All is well

entando-service:
All is well

entando-service:
All is well

PR #174, opened 2017-06-09
bug fix page tree grid, undefined frame position

mcarta-rt:
bug fix page tree grid, undefined frame position


entando-service:
All is well

PR #181, opened 2017-06-20
modifica property label.yes

fedelocci:


entando-service:
All is well

admin-console/src/main/java/com/agiletec/apsadmin/global-messages_it.properties
```diff
@@ -82,7 +82,7 @@ label.description=Descrizione
 label.empty.f=Vuota
 label.empty.m=Vuoto
 label.name=Nome
-label.yes=S\u00ec
+label.yes=Si
```
nicpuddu: it would be best to follow the agreed standard and use html entities for stressed vowels

entando-service:
All is well

PR #217, opened 2017-07-17
fix version

stefanopuddu:
update entando version to v4.3.2-SNAPSHOT


entando-service:

PR #218, opened 2017-07-17
fix version

stefanopuddu:
update pom to the correct version


entando-service:
All is well

entando-service:
All is well

PR #238, opened 2017-09-06
EN-26 - CN - Simplified and Traditional Chinese language options

fedelocci:
Cards

- EN-50
- EN-51

### DIFF
--- a/engine/src/main/java/com/agiletec/aps/system/services/lang/ISO_639 -1_langs.xml
+++ b/engine/src/main/java/com/agiletec/aps/system/services/lang/ISO_639 -1_langs.xml
@@ -550,7 +550,11 @@
 	</Lang>
 	<Lang>
 		<code>zh</code>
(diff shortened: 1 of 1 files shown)


entando-service:
All is well

engine/src/main/java/com/agiletec/aps/system/services/lang/ISO_639 -1_langs.xml
```diff
@@ -553,6 +553,10 @@
 		<descr>Chinese</descr>
 	</Lang>
 	<Lang>
+		<code>zhs</code>
+		<descr>Chinese Simplified</descr>
+	</Lang>
```
nicpuddu: it is Simplified Chinese, not Chinese Simplified

we also need to amend the other one to Traditional Chinese

entando-service:
All is well

entando-service:
All is well

PR #239, opened 2017-09-06
EN-61 CN - Highlight the page model Frame when dragging a widget over it

fedelocci:
Card EN-61 CN - Highlight the page model Frame when dragging a widget over it


entando-service:
All is well

PR #258, opened 2017-09-20
Fix bug labels spostamento attributi

natsuma:


entando-service:

PR #259, opened 2017-09-22
fix issue datepicker for attribute number

fedelocci:


entando-service:
All is well

entando-service:
All is well

PR #260, opened 2017-09-25
Revert "reserved page fix"

nicpuddu:
Reverts entando/entando-core#256


entando-service:

PR #284, opened 2018-01-24
Upgrade category management services

eugeniosant:
Upgrade category management (with cache wrapper) to facilitate insert of a DataGrid layer (realized by Infinispan connector into components project).

### DIFF
--- a/engine/src/main/java/com/agiletec/aps/system/services/category/CategoryManager.java
+++ b/engine/src/main/java/com/agiletec/aps/system/services/category/CategoryManager.java
@@ -41,8 +41,6 @@ public class CategoryManager extends AbstractService implements ICategoryManager
 
 	private static final Logger _logger = LoggerFactory.getLogger(CategoryManager.class);
 
-	public static final String RELOAD_CATEGORY_REFERENCES_THREAD_NAME_PREFIX = "RELOAD_CATEGORY_REFERENCES_";
-
 	private ILangManager _langManager;
 
 	private ICategoryManagerCacheWrapper _cacheWrapper;
--- a/engine/src/main/java/com/agiletec/aps/system/services/category/CategoryManager.java
+++ b/engine/src/main/java/com/agiletec/aps/system/services/category/CategoryManager.java
@@ -28,8 +28,10 @@ import org.springframework.beans.factory.ListableBeanFactory;
 import com.agiletec.aps.system.common.AbstractService;
 import com.agiletec.aps.system.common.tree.ITreeNode;
 import com.agiletec.aps.system.exception.ApsSystemException;
+import com.agiletec.aps.system.services.category.cache.ICategoryManagerCacheWrapper;
 import com.agiletec.aps.system.services.lang.ILangManager;
 import com.agiletec.aps.system.services.lang.Lang;
+import com.agiletec.aps.system.services.page.cache.IPageManagerCacheWrapper;
 import com.agiletec.aps.util.ApsProperties;
 import com.agiletec.aps.util.DateConverter;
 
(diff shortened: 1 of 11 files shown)


entando-service:
All is well

engine/src/main/java/com/agiletec/aps/system/services/category/CategoryManager.java
```diff
@@ -42,63 +41,22 @@
 
 	private static final Logger _logger = LoggerFactory.getLogger(CategoryManager.class);
 
+	public static final String RELOAD_CATEGORY_REFERENCES_THREAD_NAME_PREFIX = "RELOAD_CATEGORY_REFERENCES_";
+
```
joewhite101: Is this used anywhere else as a constant to add or remove from Cache? If so do you think it is worth creating a static CacheConstants file or a general Constants file with these?

If its not used elsewhere make private?

engine/src/main/java/com/agiletec/aps/system/services/category/CategoryManager.java
```diff
@@ -126,16 +84,17 @@ public void addCategory(Category category) throws ApsSystemException {
...
 			try {
 				this.getCategoryDAO().deleteCategory(code);
+				this.getCacheWrapper().deleteCategory(code);
 			} catch (Throwable t) {
 				_logger.error("Error detected while removing the category {}", code, t);
 				throw new ApsSystemException("Error detected while removing a category", t);
 			}
 		}
-		this.loadCategories();
+		this.initCache();
```
joewhite101: what does the initCache() call do after a delete? Looks like the cache us updated in the try{} block above.

engine/src/main/java/com/agiletec/aps/system/services/category/CategoryManager.java
```diff
@@ -152,16 +111,7 @@ public void updateCategory(Category category) throws ApsSystemException {
...
-	}
-
-	/**
-	 * Restituisce la mappa delle categorie, indicizzate per codice.
-	 *
-	 * @return La mappa delle categorie.
-	 */
-	private Map<String, Category> getCategories() {
-		return this._categories;
+		this.initCache();
```
joewhite101: Same question as above

engine/src/main/java/com/agiletec/aps/system/services/category/cache/CategoryManagerCacheWrapper.java
```diff
@@ -0,0 +1,151 @@
...
+		if (root == null) {
+			throw new ApsSystemException("Error found in the category tree: undefined root");
+		}
+		this.insertObjectsOnCache(cache, root, categoryMap);
+	}
+
+	protected void releaseCachedObjects(Cache cache) {
+		List<String> codes = (List<String>) this.get(cache, CATEGORY_CODES_CACHE_NAME, List.class);
+		if (null != codes) {
+			for (int i = 0; i < codes.size(); i++) {
```
joewhite101: Is this loop needed? Looks like we could just set CATEGORY_CODES_CACHE_NAME to a new List<String> and the old stuff would be dereferenced. Or is there other stuff in there?

engine/src/main/java/com/agiletec/aps/system/services/category/CategoryManager.java
eugeniosant: It is used only in the CategoryAction (for the administration area) to check how many categories are changed compared to the total (to generate a progress bar that I do not see has been implemented in front-end).
It is not a constant managed by the cache. I do not know if it is appropriate to move it in a constant... what do you think? Maybe we can move it into the interface of the CategoryManager...

engine/src/main/java/com/agiletec/aps/system/services/category/CategoryManager.java
eugeniosant: The initCache() rebuilds all the categories because some categories may still have references with the deleted category (in the property "childrenCodes" or in the "parent").

engine/src/main/java/com/agiletec/aps/system/services/category/CategoryManager.java
joewhite101: The Interface seems like a better home to me than the implementation. It looks like the other interfaces have the constants so it would be consistent.

engine/src/main/java/com/agiletec/aps/system/services/category/CategoryManager.java
joewhite101: Ok makes sense. In the future (not this PR) do you think we will need to make those references concrete to the instance so they can be updated directly without a full cache rebuild?

engine/src/main/java/com/agiletec/aps/system/services/category/cache/CategoryManagerCacheWrapper.java
eugeniosant: It been used to eliminate all the categories, to avoid that they remain categories with dirty references (see the case of the deletion of single category)

entando-service:
All is well

PR #287, opened 2018-01-25
Upgrade group manager

eugeniosant:
Upgrade of group manager (with cache wrapper) to facilitate insert of a DataGrid layer (realized by Infinispan connector into components project).

### DIFF
--- a/engine/src/main/java/com/agiletec/aps/system/services/group/cache/GroupManagerCacheWrapper.java
+++ b/engine/src/main/java/com/agiletec/aps/system/services/group/cache/GroupManagerCacheWrapper.java
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2018-Present Entando Inc. (http://www.entando.com) All rights reserved.
+ *
+ * This library is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation; either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * This library is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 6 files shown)


entando-service:
All is well

engine/src/main/java/com/agiletec/aps/system/services/group/cache/GroupManagerCacheWrapper.java
```diff
@@ -0,0 +1,139 @@
...
+public class GroupManagerCacheWrapper extends AbstractCacheWrapper implements IGroupManagerCacheWrapper {
+
+	private static final Logger _logger = LoggerFactory.getLogger(GroupManagerCacheWrapper.class);
+
+	private static enum Action {
+		ADD, UPDATE, DELETE
+	}
+
+	@Override
+	public void initCache(IGroupDAO groupDAO) throws ApsSystemException {
```
joewhite101: Probably too late now but it seems like a lot of this could be in the Abstract implementation with a parameterized type. The init, release, insert and get calls are the same across all of the implementations.  Could be something like `public void initCache(ICacheableDAO<T> cacheDAO) ...` where T is an implementation of something that caches. Or there could be an ICacheable interface that marks assets as cacheable and expose some kind of update/process methods for custom logic. Then all of this boilerplate lives in the Abstract wrapper and the custom calls like `manageGroups` below could be a managed call to the implementation.

eugeniosant: It's exactly the same thing I thought a couple of days ago. I was thinking of building the next wrappers in the same way and standardizing those already created, so as to modify them (all together) from the perspective you have described.

PR #290, opened 2018-01-31
Upgrade I18nManager

stefanopuddu:
Migrate I18nManager to data grid implementation


entando-service:
All is well

engine/src/main/java/com/agiletec/aps/system/services/i18n/cache/I18nManagerCacheWrapper.java
```diff
@@ -0,0 +1,117 @@
...
+		}
+		Cache cache = this.getCache();
+		Map<String, ApsProperties> codes = this.getLabelGroups();
+		if (Action.ADD.equals(operation)) {
+			if (!codes.containsKey(key)) {
+				codes.put(key, labels);
+				cache.put(I18N_CODES_CACHE_NAME, codes);
+			}
+			cache.put(I18N_CACHE_NAME_PREFIX + key, labels);
+		} else if (Action.UPDATE.equals(operation) && codes.containsKey(key)) {
```
joewhite101: What happens if UPDATE is called but the corresponding key isn't in the cache? Right now this is a no-op which could be really confusing and could point to a cache consistency issue in a race between an update and a delete. This should probably throw an exception so that we can return to a user that the object they are updating no longer exists. Or it should behave as an UPSERT which has its own issues. Whatever the case it will need to be consistent with the other implementations.

entando-service:
All is well

entando-service:
All is well

PR #291, opened 2018-02-01
En_333 ApiCatalogManager cache

stefanopuddu:


entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/services/api/cache/ApiCatalogManagerCacheWrapper.java
```diff
@@ -0,0 +1,158 @@
...
+		try {
+			List<ApiMethod> apiGETMethods = buildApiGetMethods(resources);
+			services = apiCatalogDAO.loadServices(apiGETMethods);
+
+			for (Map.Entry<String, ApiService> entry : services.entrySet()) {
+				cache.put(APICATALOG_SERVICE_CACHE_NAME_PREFIX + entry.getKey(), entry.getValue());
+			}
+
+		} catch (Throwable t) {
+			services = new HashMap<String, ApiService>();
```
joewhite101: This is a no-op since the exception is thrown below.

engine/src/main/java/org/entando/entando/aps/system/services/api/cache/ApiCatalogManagerCacheWrapper.java
```diff
@@ -0,0 +1,158 @@
...
+			logger.error("Error loading Services definitions", t);
+			throw new ApsSystemException("Error loading Services definitions", t);
+		}
+		cache.put(APICATALOG_RESOURCES_CACHE_NAME, resources);
+		cache.put(APICATALOG_SERVICES_CACHE_NAME, services);
```
joewhite101: Is this for sure where these belong with respect to the try call? If an exception is thrown partway through the loop we would have a mix of in and out of cache objects from the services list. May not matter and sort of a fringe case but these two calls at the end maybe out of place

engine/src/main/java/org/entando/entando/aps/system/services/api/cache/ApiCatalogManagerCacheWrapper.java
```diff
@@ -0,0 +1,158 @@
...
+		for (int i = 0; i < resourceList.size(); i++) {
+			ApiResource apiResource = resourceList.get(i);
+			if (null != apiResource.getGetMethod()) {
+				apiGETMethods.add(apiResource.getGetMethod());
+			}
+		}
+		return apiGETMethods;
+	}
+
+	protected void manage(String key, ApiService service, Action operation) {
```
joewhite101: @eugeniosant I'm wondering if we are going to want some debug logging here in all of these managers. Not suggesting that we go back and change them now but we have no logging now and we might get punished if we need to troubleshoot something on these in the future. Might be worth adding some debug at least for the big calls (evict, bulk add, init etc)

entando-service:
All is well

PR #292, opened 2018-02-01
raise exception on update null entry

stefanopuddu:
as implemented in `I18nManagerCacheWrapper` now update a null entry raises an exception


entando-service:
All is well

PR #294, opened 2018-02-05
Move commons methods of cachewrapper into abstract class

eugeniosant:


entando-service:
All is well

engine/src/main/java/com/agiletec/aps/system/common/AbstractGenericCacheWrapper.java
```diff
@@ -0,0 +1,94 @@
...
+ * @param <O> The object to manage
+ */
+public abstract class AbstractGenericCacheWrapper<O extends Object> extends AbstractCacheWrapper {
+
+	protected static enum Action {
+		ADD, UPDATE, DELETE
+	}
+
+	protected void releaseCachedObjects(Cache cache) {
+		List<String> codes = (List<String>) this.get(cache, this.getCodesCacheKey(), List.class);
```
joewhite101: I think we should add some trace or debug logging at the top level to all of these calls. What do you think?

PR #296, opened 2018-02-06
add cache management for KeyGeneratorManager

stefanopuddu:


entando-service:
All is well

engine/src/main/java/com/agiletec/aps/system/services/keygenerator/KeyGeneratorManager.java
```diff
@@ -13,74 +13,67 @@
...
-		return _keyGeneratorDao;
-	}
+    /**
+     * Restituisce la chiave univoca corrente.
+     * @return La chiave univoca corrente.
+     * @throws ApsSystemException In caso di errore 
+     * nell'aggiornamento della chiave corrente.
+     */
+    public int getUniqueKeyCurrentValue() throws ApsSystemException {
+        int key = this.getCacheWrapper().getUniqueKeyCurrentValue() + 1;
```
joewhite101: Is this thread safe?

stefanopuddu: No, it's not!

I would ask @eugeniosant for a hint

eugeniosant: IMHO, we can solve this case with "synchronized" method

engine/src/main/java/com/agiletec/aps/system/services/keygenerator/KeyGeneratorManager.java
```diff
@@ -13,74 +13,67 @@
...
-	protected IKeyGeneratorDAO getKeyGeneratorDAO() {
-		return _keyGeneratorDao;
-	}
+    /**
+     * Restituisce la chiave univoca corrente.
+     * @return La chiave univoca corrente.
+     * @throws ApsSystemException In caso di errore 
+     * nell'aggiornamento della chiave corrente.
+     */
+    public synchronized int getUniqueKeyCurrentValue() throws ApsSystemException {
```
joewhite101: The update call doesn't need to be synchronized and since it is a DAO and a cache call it would be better if it were outside the block. I think it should be something like

    synchronized (SOME_PRIVATE_STRING) { 
    int key = this.getCacheWrapper().getUniqueKeyCurrentValue() + 1;
    }
    this.updateKey(key);

entando-service:
All is well

entando-service:
All is well

engine/src/main/java/com/agiletec/aps/system/services/keygenerator/KeyGeneratorManager.java
```diff
@@ -60,7 +60,10 @@ public void init() throws Exception {
      * nell'aggiornamento della chiave corrente.
      */
     public int getUniqueKeyCurrentValue() throws ApsSystemException {
-        int key = this.getCacheWrapper().getUniqueKeyCurrentValue() + 1;
+        int key;
+        synchronized (this) {
```
joewhite101: This should be synchronized on a private member string variable. Synchronizing on `this` is not necessary and can cause issues down the line. For example, in the future there is a new method and somebody synchronizes on `this `then those two sections  are on the same lock and if one of them is long we have problems. Add a `private String UNIQUE_KEY_LOCK = "UNIQUE_KEY_LOCK" and synchronize on that.

entando-service:
All is well

PR #301, opened 2018-02-08
Oltu integration refact

paddeo:


entando-service:
All is well

nicpuddu:
this PR is not valid because master was merged into the branch, so it would bring cloud native features into the RC

PR #304, opened 2018-02-09
EN-342 Upgrade content page mapper management

eugeniosant:
Upgrade of Content Page Mapper Manager to facilitate insert of Infinispan connector (into components project).


entando-service:
All is well

PR #316, opened 2018-02-21
En 803 api interface pagination and filtering

stefanopuddu:
- switch to java8
- core dependencies: add jackson, hibernate validator and springmvc
- engine dependencies: add hamcrest, jsonpath
- add GroupService
- add dto utils
- add org.entando.entando.web package with custom exceptions, annotations, handlers and rest search related beans
- add group controller and test
- add rest/messages.properties
- add spring servlet basic setup servlet-context.xml
- add global error codes constants


engine/pom.xml
```diff
@@ -327,5 +327,18 @@
...
+			<groupId>org.hamcrest</groupId>
+			<artifactId>hamcrest-all</artifactId>
+			<version>1.3</version>
+			<scope>test</scope>
+		</dependency>
+		
+		<dependency>
+			<groupId>com.jayway.jsonpath</groupId>
+			<artifactId>json-path-assert</artifactId>
+			<version>2.2.0</version>
```
joewhite101: Scope test?

engine/src/main/java/org/entando/entando/aps/system/services/group/GroupService.java
```diff
@@ -0,0 +1,137 @@
...
+
+    public void setDtoBuilder(IDtoBuilder<Group, GroupDto> dtoBuilder) {
+        this.dtoBuilder = dtoBuilder;
+    }
+
+    @Override
+    public PagedMetadata<GroupDto> getGroups(RestListRequest restListReq) {
+        try {
+
+            Arrays.stream(restListReq.getFieldSearchFilters())
```
joewhite101: Lets add a comment here on what this is doing. Especially since this is the reference implementation.

engine/src/main/java/org/entando/entando/aps/system/services/group/GroupService.java
```diff
@@ -0,0 +1,137 @@
...
+
+            SearcherDaoPaginatedResult<Group> groups = this.getGroupManager().getGroups(restListReq.getFieldSearchFilters());
+            List<GroupDto> dtoList = dtoBuilder.convert(groups.getList());
+
+            PagedMetadata<GroupDto> pagedMetadata = new PagedMetadata<>(restListReq, groups);
+            pagedMetadata.setBody(dtoList);
+
+            return pagedMetadata;
+        } catch (Throwable t) {
+            throw new RestServerError("error in search groups", t);
```
joewhite101: log message

engine/src/main/java/org/entando/entando/aps/system/services/group/GroupService.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+
+    @Override
+    public GroupDto getGroup(String groupName) {
+        Group group = this.getGroupManager().getGroup(groupName);
+        if (null == group) {
+            throw new RestRourceNotFoundException("group", groupName);
```
joewhite101: log message

engine/src/main/java/org/entando/entando/web/common/interceptor/EntandoOauth2Interceptor.java
```diff
@@ -0,0 +1,117 @@
...
+        UserDetails user = authenticationProviderManager.getUser(username);
+        if (user != null) {
+            logger.info("User {} requesting resource that requires {} permission ", username, permission);
+            if (StringUtils.isNotBlank(permission)) {
+                if (!authorizationManager.isAuthOnPermission(user, permission)) {
+                    throw new EntandoAuthorizationException(null, request, username);
+                }
+            }
+        } else {
+            logger.info("User {} not found ", username);
```
joewhite101: Are we sure this should be a no-op?

engine/src/main/java/org/entando/entando/web/common/interceptor/EntandoOauth2Interceptor.java
```diff
@@ -0,0 +1,117 @@
...
+            String username = token.getClientId();
+            this.checkAuthorization(username, permission, request);
+
+        } catch (OAuthSystemException | ApsSystemException | OAuthProblemException ex) {
+            logger.error("System exception {}", ex.getMessage());
+            throw new RestServerError("error parsing OAuth parameters", ex);
+        }
+    }
+
+    protected void checkAuthorization(String username, String permission, HttpServletRequest request) throws ApsSystemException {
```
joewhite101: It seems odd to me that this method doesn't have a return type. Are we relying on exceptions for all auth flow control? I get why but the method no-ops if the user isn't found which is the same as "authorized". So if i made up a fake user name i get to be superuser.

entando-service:
All is well

engine/src/main/java/org/entando/entando/web/common/interceptor/EntandoOauth2Interceptor.java
```diff
@@ -0,0 +1,117 @@
...
+
+            final OAuth2Token token = oAuth2TokenManager.getApiOAuth2Token(accessToken);
+            this.validateToken(request, accessToken, token);
+
+            String username = token.getClientId();
+            this.checkAuthorization(username, permission, request);
+
+        } catch (OAuthSystemException | ApsSystemException | OAuthProblemException ex) {
+            logger.error("System exception {}", ex.getMessage());
+            throw new RestServerError("error parsing OAuth parameters", ex);
```
joewhite101: This doesn't match up with the checkAuthorization messaging. If a User  isn't authorized an exception is thrown but here we are saying "error parsing parameters"

engine/src/main/java/org/entando/entando/web/common/model/Filter.java
```diff
@@ -0,0 +1,113 @@
...
+        return value;
+    }
+
+    public void setValue(String value) {
+        this.value = value;
+    }
+
+
+    public Filter() {
+        //
```
joewhite101: // remove

engine/src/main/java/org/entando/entando/web/common/model/Filter.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @SuppressWarnings("rawtypes")
+    public FieldSearchFilter getFieldSearchFilter() {
+        FieldSearchFilter filter = new FieldSearchFilter(StringEscapeUtils.escapeSql(this.getAttributeName()), StringEscapeUtils.escapeSql(this.getValue()), true, LikeOptionType.COMPLETE);
+        return filter;
+    }
+
+    @Override
+    public int hashCode() {
```
joewhite101: Do we need this?

engine/src/main/java/org/entando/entando/web/common/model/PagedMetadata.java
```diff
@@ -0,0 +1,90 @@
...
+import com.agiletec.aps.system.common.model.dao.SearcherDaoPaginatedResult;
+
+public class PagedMetadata<T> {
+
+    private int page;
+    private int size;
+    private int last;
+    private int count;
+
+    //@JsonIgnore
```
joewhite101: Remove

engine/src/main/java/org/entando/entando/web/common/model/PagedMetadata.java
```diff
@@ -0,0 +1,90 @@
...
+    private List<T> body;
+
+    public PagedMetadata() {
+        //
```
joewhite101: // remove

engine/src/main/java/org/entando/entando/web/common/model/RestError.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+package org.entando.entando.web.common.model;
+
+public class RestError {
+
+	private String code;
+	private String message;
+
+	public RestError() {
+		//
```
joewhite101: // remove

engine/src/main/java/org/entando/entando/web/common/model/RestListRequest.java
```diff
@@ -0,0 +1,165 @@
...
+
+public class RestListRequest {
+
+    private String sort;
+    private String direction;
+
+    private Integer pageNum = 0;
+    private Integer pageSize = 5;
+
+    private Filter[] filter;
```
joewhite101: If possible lets make this plural. `filters` May have to us an annotatation for the json conversion.

engine/src/main/java/org/entando/entando/web/common/model/RestListRequest.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    public void addFilter(Filter filter) {
+        this.filter = ArrayUtils.add(this.filter, filter);
+    }
+
+    @SuppressWarnings("rawtypes")
+    public FieldSearchFilter[] getFieldSearchFilters() {
+        FieldSearchFilter[] filters = new FieldSearchFilter[0];
+        if (null != filter && filter.length > 0) {
```
joewhite101: Going forward, lets make new code List<> and use streams. Then if we absolutely have to return an array we can convert it right before return.

engine/src/main/java/org/entando/entando/web/common/model/RestListRequest.java
```diff
@@ -0,0 +1,165 @@
...
+        int page = this.getPageNum();
+        if (null == this.getPageNum() || this.getPageNum() == 0) {
+            return 0;
+        }
+        return this.getPageSize() * page;
+    }
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
```
joewhite101: We need this?

engine/src/main/java/org/entando/entando/web/common/model/RestResponse.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.List;
+
+public class RestResponse {
+
+	private Object payload;
+	private List<RestError> errors = new ArrayList<>();
+	private Object metadata;
+
+	public RestResponse() {
+		//
```
joewhite101: // remove

engine/src/main/java/org/entando/entando/web/group/model/GroupPutRequest.java
```diff
@@ -0,0 +1,45 @@
...
+package org.entando.entando.web.group.model;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * group put payload 
+ *
+ */
+public class GroupPutRequest {
+
```
joewhite101: Is this Put specific? What is the structural difference between a Put and a Post? No the verb names in the object names.

engine/src/main/java/org/entando/entando/web/group/model/GroupRequest.java
```diff
@@ -0,0 +1,33 @@
+package org.entando.entando.web.group.model;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * group add payload 
+ *
+ */
+public class GroupRequest extends GroupPutRequest {
+
```
joewhite101: I am confused by the need for extension here. Can't be just one object?

engine/src/test/java/com/agiletec/aps/AllTests.java
```diff
@@ -191,6 +191,10 @@ public static Test suite() {
 
 		suite.addTestSuite(TestSearchEngineManager.class);
 
+        //$JUnit-BEGIN$
```
joewhite101: What do these comments do?

engine/src/main/java/org/entando/entando/web/common/interceptor/EntandoOauth2Interceptor.java
paddeo: Yes. It's the case when an access token represents an API consumer but not an Entando user, so we should not deny access.

engine/src/main/java/org/entando/entando/web/common/interceptor/EntandoOauth2Interceptor.java
paddeo: you can not get to this point if you have not obtained a valid access token, ie if you are neither an authorized API consumer nor an authenticated Entando user

engine/src/main/java/org/entando/entando/web/common/interceptor/EntandoOauth2Interceptor.java
paddeo: Those exceptions (but not ApsSystemException) depend on the library we are using (Oltu): if you are wrong adding the token to the request, the library class responsible for request management raises one of them

engine/src/main/java/org/entando/entando/web/common/model/Filter.java
stefanopuddu: I need this in my tests to compare `RestListRequest`s

engine/src/main/java/org/entando/entando/web/common/model/PagedMetadata.java
stefanopuddu: thanks for pointing me out. This annotation must be active

engine/src/main/java/org/entando/entando/web/group/model/GroupRequest.java
stefanopuddu: This is all about inheritance of  validation annotations:
in this case the put demands one required field, and the post demands the same field, with the same validation rules plus another filed...

We can have of course two independent objects

joewhite101: What is specific? The payloads for PUT and POST should be identical except for whatever is the primary key field. And that is available in the URI for POST right?

entando-service:
All is well

entando-service:
All is well

PR #320, opened 2018-02-28
fix mismatch code/name

stefanopuddu:
fixes issue EN-859


entando-service:
All is well

engine/src/main/java/org/entando/entando/web/group/GroupController.java
```diff
@@ -33,13 +33,15 @@
...
 import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
 import org.springframework.validation.BindingResult;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RestController;
 
+@Controller
```
joewhite101: Do we need both Controller and RestController?

entando-service:
All is well

PR #339, opened 2018-03-13
cors filter added

paddeo:


engine/src/main/java/org/entando/entando/aps/servlet/CORSFilter.java
```diff
@@ -0,0 +1,36 @@
...
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain) throws ServletException, IOException {
+        response.addHeader("Access-Control-Allow-Origin", "*");
+
+        if (request.getHeader("Access-Control-Request-Method") != null && "OPTIONS".equals(request.getMethod())) {
+            logger.trace("Sending Header....");
+            // CORS "pre-flight" request
+            response.addHeader("Access-Control-Allow-Methods", "GET, POST, PUT, DELETE");
```
nicpuddu: the methods should always be included and not only if options is being submitted.
also options should be among the allowed methods.

paddeo: Got it.

entando-service:
All is well

engine/src/test/java/org/entando/entando/web/system/ReloadConfigurationControllerTest.java
```diff
@@ -10,25 +10,27 @@
...
 
         UserDetails user = new OAuth2TestUtils.UserBuilder("jack_bauer", "0x24").grantedToRoleAdmin().build();
         String accessToken = mockOAuthInterceptor(user);
         ResultActions result = mockMvc
-                                      .perform(post("/reloadConfiguration")
-                                                                           .header("Authorization", "Bearer " + accessToken));
+                .perform(post("/reloadConfiguration")
+                        .header("Authorization", "Bearer " + accessToken));
         result.andExpect(status().isOk());
+        result.andExpect(header().string("Access-Control-Allow-Origin", "*"));
```
joewhite101: Please  add a comment about what is being tested here so that if this breaks we know why it is here and what it is testing. 6 months from now if this fails we'll be really confused

entando-service:
All is well

entando-service:
All is well

PR #351, opened 2018-03-19
page children fix

paddeo:


entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/services/page/PageService.java
```diff
@@ -129,9 +129,11 @@ public void setWidgetTypeManager(IWidgetTypeManager widgetTypeManager) {
...
         IPage parent = this.getPageManager().getDraftPage(parentCode);
         Optional.ofNullable(parent).ifPresent(root -> Optional.ofNullable(parent.getChildrenCodes()).ifPresent(children -> Arrays.asList(children).forEach(childCode -> {
-            IPage child = this.getPageManager().getOnlinePage(childCode) != null
-                    ? this.getPageManager().getOnlinePage(childCode) : this.getPageManager().getDraftPage(childCode);
-            res.add(dtoBuilder.convert(child));
+            IPage childO = this.getPageManager().getOnlinePage(childCode),
+                    childD = this.getPageManager().getDraftPage(childCode);
+            PageDto child = childO != null ? dtoBuilder.convert(childO) : dtoBuilder.convert(childD);
+            child.setChildren(Arrays.asList(childD.getChildrenCodes()));
+            res.add(child);
```
joewhite101: Why do we always return the children of the draft page?

paddeo: Because everytime in page tree we need all the children of a page, not only published ones.

PR #358, opened 2018-03-20
En 871 user authorities endpoints

paddeo:


entando-service:

entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/services/user/UserService.java
```diff
@@ -0,0 +1,83 @@
...
+    public List<UserAuthorityDto> addUserAuthorities(String username, UserAuthoritiesRequest request) {
+        List<UserAuthorityDto> authorizations = new ArrayList<>();
+        try {
+            request.forEach(authorization
+                    -> {
+                try {
+                    this.getAuthorizationManager().getUserAuthorizations(username);
+                    this.getAuthorizationManager().addUserAuthorization(username, authorization.getGroup(), authorization.getRole());
+                } catch (ApsSystemException ex) {
+                    throw new RuntimeException(ex);
```
joewhite101: Probably should throw the same exception as below. Also, add logging.
throw new RestServerError("Error in delete authorities", e);

engine/src/main/java/org/entando/entando/web/user/model/UserRequest.java
```diff
@@ -0,0 +1,22 @@
...
+ * FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ */
+package org.entando.entando.web.user.model;
+
+/**
+ *
+ * @author paddeo
+ */
+public class UserRequest {
```
joewhite101: What does this get used for? Can we drop it?

engine/src/main/java/org/entando/entando/aps/system/services/user/UserService.java
paddeo: Yes, I'll fix this.

engine/src/main/java/org/entando/entando/web/user/model/UserRequest.java
paddeo: I created this object for mistake but, since I need it for users endpoints, I decided to keep it. But I'll drop it and re-create inside my "users endpoints" branch.

entando-service:
All is well

entando-service:

entando-service:
All is well

PR #360, opened 2018-03-21
add EntandoMessageCodesResolver

stefanopuddu:


entando-service:
All is well

engine/src/main/java/org/entando/entando/web/common/EntandoMessageCodesResolver.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.springframework.util.StringUtils;
+import org.springframework.validation.DefaultMessageCodesResolver;
+import org.springframework.validation.MessageCodeFormatter;
+
+public class EntandoMessageCodesResolver extends DefaultMessageCodesResolver {
```
joewhite101: Please add a high level description of this in a comment. This is pretty invisible functionality so we need to make sure to describe what's going on here.

engine/src/main/java/org/entando/entando/web/label/LabelController.java
```diff
@@ -58,6 +58,7 @@ public void setLabelValidator(LabelValidator labelValidator) {
     @RestAccessControl(permission = Permission.SUPERUSER)
     @RequestMapping(method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
     public ResponseEntity<?> getLables(RestListRequest requestList) {
+        logger.debug("loading labels");
```
joewhite101: Does this requestList contain anything?

stefanopuddu: Yes, labels can be filtered by key and by value

entando-service:
All is well

PR #369, opened 2018-03-25
EN-1075 fix dataModel Delete payload

eugeniosant:


entando-service:
All is well

entando-service:
All is well

PR #388, opened 2018-03-29
En 1099 fix

eugeniosant:
Improve validation into Widget DELETE and fix null filters element into the metadata in payload


entando-service:
All is well

PR #392, opened 2018-03-30
En 995 intellisense support contentmodel

stefanopuddu:
api: add contentmodel endpoints and intellisense support


entando-service:
All is well

PR #405, opened 2018-04-09
En 1107 update error handling with req list

paddeo:


entando-service:
All is well

engine/src/main/java/org/entando/entando/web/common/model/Filter.java
```diff
@@ -20,8 +20,10 @@
 
 public class Filter {
 
+    private static final String DEFAULT_OP = "like";
```
joewhite101: This looks really brittle to me. Why do we have the default as "like"?

engine/src/main/java/org/entando/entando/web/database/validator/DatabaseValidator.java
```diff
@@ -37,4 +42,20 @@ public void validate(Object o, Errors errors) {
...
     }
 
+    @Override
+    public boolean isValidField(String fieldName) {
+        List<String> fields = new ArrayList<>();
+        fields = getAllFields(fields, DumpReportDto.class);
+        return fields.contains(fieldName);
+    }
+
+    List<String> getAllFields(List<String> fields, Class<?> type) {
```
joewhite101: This is identical to the one in the DataTypeValidator. Can we move to a utility class?

engine/src/main/java/org/entando/entando/web/guifragment/validator/GuiFragmentValidator.java
```diff
@@ -88,4 +93,19 @@ private int validateGuiCode(GuiFragmentRequestBody request, Errors errors) {
...
     }
 
+    @Override
+    public boolean isValidField(String fieldName) {
+        List<String> fields = new ArrayList<>();
+        fields = getAllFields(fields, GuiFragmentDto.class);
+        return fields.contains(fieldName);
+    }
+
+    List<String> getAllFields(List<String> fields, Class<?> type) {
```
joewhite101: Move to utility.

engine/src/main/java/org/entando/entando/web/label/LabelValidator.java
```diff
@@ -50,4 +54,11 @@ public void validateBodyName(String labelCode, LabelRequest labelRequest, Errors
             errors.rejectValue("key", ERRCODE_URINAME_MISMATCH, new String[]{labelCode, labelRequest.getKey()}, "labelRequest.key.mismatch");
         }
     }
+
+    @Override
+    public boolean isValidField(String fieldName) {
```
joewhite101: This should be done generically. Given that the reference is static it can be a parameter to the call.

engine/src/main/java/org/entando/entando/web/language/validator/LanguageValidator.java
```diff
@@ -34,4 +38,10 @@ public void validate(Object target, Errors errors) {
 
     }
 
+    @Override
+    public boolean isValidField(String fieldName) {
```
joewhite101: Move to utility.

engine/src/main/java/org/entando/entando/web/page/validator/PageValidator.java
```diff
@@ -190,4 +193,11 @@ public void validateReferences(String pageCode, PageStatusRequest pageStatusRequ
         }
     }
 
+    @Override
+    public boolean isValidField(String fieldName) {
```
joewhite101: Move to utility.

engine/src/main/java/org/entando/entando/web/pagemodel/validator/PageModelValidator.java
```diff
@@ -79,4 +81,10 @@ public void validateBodyName(String code, PageModelRequest pageModelRequest, Err
 
     }
 
+    @Override
+    public boolean isValidField(String fieldName) {
```
joewhite101: Move to utility.

engine/src/main/java/org/entando/entando/web/permission/validator/PermissionValidator.java
```diff
@@ -30,5 +33,10 @@ public void validate(Object target, Errors errors) {
 
     }
 
+    @Override
+    public boolean isValidField(String fieldName) {
```
joewhite101: Move to utility.

engine/src/main/java/org/entando/entando/web/common/model/Filter.java
paddeo: "like" operator is generally used for a search with filters, when you want more than one specific result, so I think is better than "eq" operator, as default. It's up to managers and daos to ensure that "like" works fine also with numbers and dates.

entando-service:
All is well

engine/src/main/java/org/entando/entando/web/common/model/Filter.java
joewhite101: It was "null" before though so it would fail if not specified. This seems like a really big change. @eugeniosant and @stefanopuddu what do you guys think?

entando-service:
All is well

engine/src/main/java/org/entando/entando/web/common/model/Filter.java
paddeo: you still don't need to specify a filter: the default operator is active only if you add a filter in your request. on the other side, if we remove the default operator, it always fails if we add a filter without specifying an operator.

stefanopuddu: Actually the `like` operator was implicit in the previous implementation.
Check the `getFieldSearchFilter()` method. This is where the `Filter` is (must be) converted into a working Entando filter

PR #427, opened 2018-04-18
En 1148 spring upgrade

eugeniosant:


entando-service:
All is well

engine/src/main/java/org/springframework/aop/support/AopUtils.java
```diff
@@ -0,0 +1,408 @@
...
+import org.springframework.aop.SpringProxy;
+import org.springframework.aop.TargetClassAware;
+import org.springframework.core.BridgeMethodResolver;
+import org.springframework.core.MethodIntrospector;
+import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
+import org.springframework.util.ReflectionUtils;
+
+/**
```
joewhite101: We need a REALLY detailed comment on why this is here in our codebase since it overrides something that comes from Spring and eventually could be very very confusing for a user of the platform.

joewhite101:
Retest this please. test again

engine/src/main/java/org/springframework/aop/support/AopUtils.java
eugeniosant: Done

entando-service:
All is well

entando-service:
All is well

entando-service:
All is well

pietrangelo:
retest this please

entando-service:
All is well

PR #435, opened 2018-04-20
add free pages endpoint

stefanopuddu:
add endpoint `/pages/search/group/free`

Similar to the standard search feature


entando-service:
All is well

PR #461, opened 2018-05-07
add fields on page payload

paddeo:


engine/src/main/java/org/entando/entando/aps/system/services/page/model/PageDto.java
```diff
@@ -88,6 +91,11 @@ public PageDto(IPage page) {
                 .filter(widget -> widget != null)
                 .collect(Collectors.toList())
                 .size()));
+        if (null != page.getMetadata().getUpdatedAt()) {
+            SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
```
eugeniosant: Here you can use the constants SystemConstants.API_DATE_FORMAT

paddeo: ok!

entando-service:
All is well

entando-service:
All is well

PR #466, opened 2018-05-10
pagination fix

paddeo:


entando-service:

engine/src/main/java/org/entando/entando/web/common/model/PagedMetadata.java
```diff
@@ -126,4 +134,25 @@ public void setFilters(Filter[] filters) {
...
+    public void setActualSize(int actualSize) {
+        this.actualSize = actualSize;
+    }
+
+    public void imposeLimits() {
+        if (((this.getPage() - 1) * this.getActualSize()) > this.getBody().size()) {
+            this.setBody(new ArrayList<>());
+        } else {
+            int start = ((this.getPage() - 1) * this.getActualSize());
+            int end = (this.getPage() * this.getActualSize()) <= this.getBody().size() ? (this.getPage() * this.getActualSize())
```
joewhite101: This is way way too complex. I can't follow what is happening at all. Break this up into multiple lines with variable names that tell us what is happening. Avoid the tertiary operator in chains.

entando-service:
All is well

engine/src/main/java/org/entando/entando/web/common/model/PagedMetadata.java
paddeo: I deleted a useless line (now we don't have tertiary operator in chains) and added a comment.

PR #474, opened 2018-05-18
En 1499 En 1500 upgrade dataType refresh endpoints

eugeniosant:
Upgraded also profileType refresh endpoints


entando-service:
All is well

entando-service:
All is well

PR #484, opened 2018-05-26
update test code to include new required properties so that downstrea…

joewhite101:
…m controller tests can function


entando-service:
All is well

PR #524, opened 2018-07-06
En 1685 fix widget put endpoint

eugeniosant:


entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/services/widgettype/WidgetService.java
```diff
@@ -234,6 +242,8 @@ public WidgetDto updateWidget(String widgetCode, WidgetRequest widgetRequest) {
                 }
             }
             this.addFragments(widgetDto);
+        } catch (ValidationGenericException vge) {
```
joewhite101: Log this. No exceptions without logging.

entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/services/widgettype/WidgetService.java
eugeniosant: Done!

PR #542, opened 2018-08-06
EN-1839-434-fix on pagination for user roles

DMinnai:


DMinnai:
Added missing <s:form role="search" > that was missing on jsp file

eugeniosant:
modification already in place

PR #545, opened 2018-08-08
En 1866 fix content list cache key

eugeniosant:
The problem was found in the content list widget.
when in a content list we create a filter directly on the jsp, and this filter is dynamic (for example based on the attributes of the current user), the key with which the object is cached is the same and the resulting list is always the same.
This PR fixes this problem

### DIFF
--- a/engine/src/main/java/com/agiletec/plugins/jacms/aps/system/services/content/widget/ContentListHelper.java
+++ b/engine/src/main/java/com/agiletec/plugins/jacms/aps/system/services/content/widget/ContentListHelper.java
@@ -41,8 +41,10 @@ import com.agiletec.aps.util.ApsProperties;
 import com.agiletec.plugins.jacms.aps.system.services.content.helper.BaseContentListHelper;
 import com.agiletec.plugins.jacms.aps.system.services.content.helper.IContentListFilterBean;
 import com.agiletec.plugins.jacms.aps.system.services.content.widget.util.FilterUtils;
+import java.util.Arrays;
 import java.util.Collections;
 import org.apache.commons.codec.digest.DigestUtils;
+import org.apache.commons.lang3.ArrayUtils;
 import org.springframework.cache.annotation.Cacheable;
 
 /**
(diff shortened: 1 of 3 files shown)


entando-service:
All is well

entando-service:
All is well

engine/src/main/java/com/agiletec/plugins/jacms/aps/system/services/content/widget/ContentListHelper.java
```diff
@@ -252,23 +254,18 @@ protected boolean extractOrCategoryFilterClause(ApsProperties config) {
...
         String widgetFilters = widgetParams.getProperty(widgetParamName);
         EntitySearchFilter[] filters = this.getFilters(bean.getContentType(), widgetFilters, reqCtx);
         if (null == filters) {
-            return;
-        }
-        for (EntitySearchFilter filter : filters) {
-            bean.addFilter(filter);
+            return bean.getFilters();
         }
+        EntitySearchFilter[] filtersToReturn = (null != bean.getFilters()) ? Arrays.copyOf(bean.getFilters(), bean.getFilters().length) : new EntitySearchFilter[0];
```
joewhite101: Way too complex for a trinary operator. Tough to see what is happening here. Please pul l this apart.

engine/src/main/java/com/agiletec/plugins/jacms/aps/system/services/content/widget/ContentListHelper.java
```diff
@@ -252,23 +254,18 @@ protected boolean extractOrCategoryFilterClause(ApsProperties config) {
...
-    @Deprecated
-    protected void addShowletFilters(IContentListTagBean bean, ApsProperties showletParams, String showletParamName, RequestContext reqCtx) {
-        this.addWidgetFilters(bean, showletParams, showletParamName, reqCtx);
-    }
-
-    protected void addWidgetFilters(IContentListTagBean bean, ApsProperties widgetParams, String widgetParamName, RequestContext reqCtx) {
+    protected EntitySearchFilter[] createWidgetFilters(IContentListTagBean bean, ApsProperties widgetParams, String widgetParamName, RequestContext reqCtx) {
         if (null == widgetParams) {
-            return;
+            return bean.getFilters();
```
joewhite101: This looks like a pretty big change possibly. Given that nothing would get set in the old call. Is this covered by a test and do we know that the contract is correct if `widgetParams` is null?

engine/src/main/java/com/agiletec/plugins/jacms/aps/system/services/content/widget/ContentListHelper.java
eugeniosant: Removed trinary operator

engine/src/main/java/com/agiletec/plugins/jacms/aps/system/services/content/widget/ContentListHelper.java
eugeniosant: The signature of the method has changed; before the modification the method added the filters into the "bean" object (erroneously changing the key with which the list was searched in cache), now it creates and returns a list of filters without altering the "bean" object.
BTW, I added other junit tests in order to cover other cases.

PR #550, opened 2018-08-28
EN-1858 Modify content attribute to store and show image metadata in content (version 4.3.4)

eugeniosant:


entando-service:

entando-service:
All is well

admin-console/src/main/java/com/agiletec/plugins/jacms/apsadmin/content/ContentAdminAction.java
```diff
@@ -1,126 +1,238 @@
...
+     * referenze tra Contenuti e Pagine, Risorse, Gruppi e Contenuti stessi in
+     * apposita tabella (elementi a servizio di controlli di autorizzazione e
+     * validazione).
+     *
+     * @return Il codice del risultato dell'azione.
+     */
+    public String reloadContentsReference() {
+        try {
+            String result = this.configSystemParams();
+            if (!result.equals(SUCCESS)) {
```
joewhite101: Just to check is this return correct? If the first one fails you don't want the contentManager call?

engine/src/main/db/update/scriptPort.sql
```diff
@@ -1,5 +1,14 @@
+-- EN-1748
 ALTER TABLE pages_metadata_draft ADD COLUMN groupcode character varying(30);
 ALTER TABLE pages_metadata_online ADD COLUMN groupcode character varying(30);
 UPDATE pages_metadata_online SET groupcode = (SELECT pages.groupcode FROM pages WHERE pages_metadata_online.code = pages.code);
 UPDATE pages_metadata_draft SET groupcode = (SELECT pages.groupcode FROM pages WHERE pages_metadata_draft.code = pages.code);
 ALTER TABLE pages DROP groupcode;
+
+-- EN-1858
+INSERT INTO sysconfig (version, item, descr, config) VALUES ('production', 'jacms_resourceMetadataMapping', 'Mapping between resource Metadata and resource attribute fields', '<mapping>
```
joewhite101: Will this load automatically for CDP or do we have something else to do when we give them 4.3.4?

admin-console/src/main/java/com/agiletec/plugins/jacms/apsadmin/content/ContentAdminAction.java
eugeniosant: yes, it's only to check the first call; you could eliminate this check (very difficult for it to fail), I put it to be sure

joewhite101:
https://confluence.entando.org/display/PP/Porting+from+5.0.0+to+5.1.0

PR #576, opened 2018-09-06
Sonar en 1853 - Add maven dependency-check plugin 

pietrangelo:


entando-service:
All is well

entando-service:
All is well

PR #625, opened 2018-10-05
EN-1939 & EN-266 fixing delete action 

otarza:
Fixing delete action for not-yet initialized cropper items.


entando-service:

PR #626, opened 2018-10-05
EN-1939 & EN-266 fixing inconsistent description input values for cropped images.

otarza:


entando-service:

PR #629, opened 2018-10-05
EN-1939 & EN-266 Fixing inconsistent ID generation.

otarza:


entando-service:

PR #634, opened 2018-10-11
EN-2048 Improvement of page tree result

eugeniosant:
The improvement regard the opening of the root node at the first access to the ui that shows the search result


entando-service:

PR #637, opened 2018-10-19
En 2130 swagger

rzsouza:
Creating a PR for early feedback. In this version swagger will be autoconfigured in a new generated app. It will be available in APP_BASE_URL/api/swagger-ui.html  (you can use client_id = swagger and secret = swaggerswagger)

### DIFF
--- a/engine/src/main/java/org/entando/entando/aps/servlet/AuthEndpointServlet.java
+++ b/engine/src/main/java/org/entando/entando/aps/servlet/AuthEndpointServlet.java
@@ -130,7 +130,7 @@ public class AuthEndpointServlet extends HttpServlet {
                     throw OAuthUtils.handleOAuthProblemException("Invalid clientId");
                 } else if (clientDetail.getExpirationDate().getTime() < System.currentTimeMillis()) {
                     throw OAuthUtils.handleOAuthProblemException("ClientId is expired");
-                } else if (!clientDetail.getCallbackUrl().equals(oauthRequest.getRedirectURI())) {
+                } else if (!oauthRequest.getRedirectURI().endsWith(clientDetail.getCallbackUrl())) {
                     throw OAuthUtils.handleOAuthProblemException("Invalid redirectUri");
                 }
 
--- a/engine/src/main/java/org/entando/entando/aps/servlet/AuthEndpointServlet.java
+++ b/engine/src/main/java/org/entando/entando/aps/servlet/AuthEndpointServlet.java
@@ -130,7 +130,7 @@ public class AuthEndpointServlet extends HttpServlet {
                     throw OAuthUtils.handleOAuthProblemException("Invalid clientId");
                 } else if (clientDetail.getExpirationDate().getTime() < System.currentTimeMillis()) {
                     throw OAuthUtils.handleOAuthProblemException("ClientId is expired");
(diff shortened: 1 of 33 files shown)


engine/src/main/resources/sql/entandoServ_data_production.sql
```diff
@@ -40,6 +40,7 @@ INSERT INTO authuserprofiles (username, profiletype, profilexml, publicprofile)
 <profile id="admin" typecode="PFL" typedescr="Default user profile"><descr /><groups /><categories /><attributes><attribute name="fullname" attributetype="Monotext" /><attribute name="email" attributetype="Monotext" /></attributes></profile>
 ', 0);
 
+INSERT INTO api_oauth_consumers (consumerkey, consumersecret, name, description, callbackurl, scope, authorizedgranttypes, expirationdate, issueddate) VALUES ('swagger', 'swaggerswagger', 'Swagger', 'Swagger', '/api/webjars/springfox-swagger-ui/oauth2-redirect.html', 'global', 'authorization_code', '2028-10-10 00:00:00', '2008-10-10 00:00:00');
```
rzsouza: In this part it would be better to add a complete URL including APP_BASE_URL. I was considering adding a SpringInitializer to add the value in the DB. I'm open to suggestions about smarter ways of doing this.

engine/src/main/java/org/entando/entando/aps/servlet/AuthEndpointServlet.java
```diff
@@ -130,7 +130,7 @@ private boolean validateClient(final OAuthAuthzRequest oauthRequest, HttpServlet
                     throw OAuthUtils.handleOAuthProblemException("Invalid clientId");
                 } else if (clientDetail.getExpirationDate().getTime() < System.currentTimeMillis()) {
                     throw OAuthUtils.handleOAuthProblemException("ClientId is expired");
-                } else if (!clientDetail.getCallbackUrl().equals(oauthRequest.getRedirectURI())) {
+                } else if (!oauthRequest.getRedirectURI().endsWith(clientDetail.getCallbackUrl())) {
```
rzsouza: This needs to be changed back, after we decide in the best approach to add the full URL in DB, see my other comment

entando-service:
All is well

engine/src/main/resources/sql/entandoServ_data_production.sql
joewhite101: @eugeniosant does this fit with the direction you're taking on the spring security update?

joewhite101:
@rzsouza This looks good to me. 

@eugeniosant Please take a look at this one.

engine/src/main/resources/sql/entandoServ_data_production.sql
eugeniosant: if you want to fix the "relative" redirect_url after the initialization you can extend the init method of ConsumerManager to update the value of this url with the "base_url" given by the URLManager.
This update can be done only for whagger consumer (and for this reason this key could be defined in the system constants).

engine/src/main/java/org/entando/entando/aps/servlet/AuthEndpointServlet.java
eugeniosant: look at the previous note

entando-service:
All is well

entando-service:
All is well

PR #638, opened 2018-10-19
EN-2136 Fixed the multipart upload bug - list of files out of sync

rzsouza:


joewhite101:
Looks good. We'll want to add tests when this gets targeted at master

PR #644, opened 2018-10-23
EN-2132 Removed join button for root category; Renamed All category into Root

zonia3000:


entando-service:
All is well

eugeniosant:
The ui can be improved removing the delete function (into kebab button) on nodes that have children

PR #659, opened 2018-10-26
EN-2128 Reverted back to 1ea0a39 and reapplied crop-editor/multi upload/drag-drop changes

otarza:


entando-service:
All is well

PR #669, opened 2018-10-30
EN-2121 fixes broken file download in file browser [master]

otarza:
https://jira.entando.org/browse/EN-2121


entando-service:
All is well

PR #675, opened 2018-10-31
EN-2159 (4.3.6) Fixed the page tree move behavior on drag and drop

zonia3000:
Improved also error messages.


entando-service:
All is well

eugeniosant:
The js function is ok. 
If you try to move a free node inside a free node it's ok, if you try to move a reserved page into a free node the the system returns an error message (and it it should be permitted), reserved page into other reserved node (with same group) is ok.
Public page inside a no public page returns a js error (parsererror 200 SyntaxError: JSON.parse: unexpected character at line 2 column 1 of the JSON data) and broke the system.

entando-service:
All is well

eugeniosant:
Verified! Seems good

PR #678, opened 2018-10-31
EN-2120 Fixed category tree not expanding properly. Aligned code with…

rzsouza:
EN-2120 Fixed category tree not expanding properly. Aligned code with version 4.3.6, removed dependency on entando-component page


entando-service:
All is well

PR #692, opened 2018-11-06
En 2168 [436] content list bugs + columns composition and position states saved on reload

DMinnai:


entando-service:
All is well

PR #696, opened 2018-11-06
Fix for Swagger/Spring Security

rzsouza:
I changed the grant type to password. Everything else is the same. I'll need to update the instructions to use swagger (need to use also user/password authentication with clientId/secret)


entando-service:
All is well

PR #706, opened 2018-11-07
EN-2184 

rzsouza:
Added check to avoid deletion for admin user.

Code is working now but might need improvements in the error handling for FE

### DIFF
--- a/engine/src/main/java/org/entando/entando/web/user/UserController.java
+++ b/engine/src/main/java/org/entando/entando/web/user/UserController.java
@@ -37,7 +37,6 @@ import org.springframework.http.ResponseEntity;
 import org.springframework.validation.BindingResult;
 import org.springframework.validation.DataBinder;
 import org.springframework.validation.MapBindingResult;
-import org.springframework.validation.ObjectError;
 import org.springframework.validation.annotation.Validated;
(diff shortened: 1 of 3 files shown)


rzsouza:
Last commit added a specific validation exception that can be used by FE to display specific error

entando-service:
All is well

engine/src/main/java/org/entando/entando/web/user/UserController.java
```diff
@@ -163,6 +165,15 @@ public void setUserValidator(UserValidator userValidator) {
         return new ResponseEntity<>(new RestResponse(result), HttpStatus.OK);
     }
 
+    private BindingResult createBindingResult(@PathVariable String username) {
+        Map<String, String> map = new HashMap<>();
+        map.put("username", username);
+        BindingResult bindingResult = new MapBindingResult(map, "username");
+        ObjectError error = new ObjectError("username", "Sorry. You can't delete the administrator user");
```
eugeniosant: you have to use message.properties file for text messages

rzsouza: Fixed

entando-service:
All is well

PR #709, opened 2018-11-08
EN-2048 Added selection of the node on 'View In Tree' action

zonia3000:


entando-service:
All is well

entando-service:
All is well

PR #713, opened 2018-11-09
EN-1639 Update Entity Attribute payload element (master)

eugeniosant:


entando-service:
All is well

entando-service:
All is well

PR #715, opened 2018-11-09
EN-2132 (5.0.1) Fixed category tree behavior for classic nodes style in Content edit

zonia3000:
This pull request is associated with [PR #386 in entando-components](https://github.com/entando/entando-components/pull/386).
If you want to test this changes you need to use also that changes.


entando-service:
All is well

PR #720, opened 2018-11-12
Upgrade of WebSecurityAdapter

eugeniosant:
disabled X-Frame-Options response header in Spring Security


entando-service:
All is well

PR #722, opened 2018-11-13
adding missing info strings for data type entry tooltips

jeffgo10:
data type entry labels for tooltip added


entando-service:
All is well

PR #723, opened 2018-11-13
EN-2184 Changed User Validator to not allow delete admin

rzsouza:
Same as 5.0.1

### DIFF
--- a/engine/src/main/java/org/entando/entando/web/user/UserController.java
+++ b/engine/src/main/java/org/entando/entando/web/user/UserController.java
@@ -17,37 +17,27 @@ import com.agiletec.aps.system.exception.ApsSystemException;
 import com.agiletec.aps.system.services.role.Permission;
 import com.agiletec.aps.system.services.user.UserDetails;
 import org.entando.entando.aps.system.services.user.IUserService;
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/entando/entando/web/user/UserController.java
```diff
@@ -150,12 +140,25 @@ public void setUserValidator(UserValidator userValidator) {
...
+            throw new ValidationGenericException(createDeleteAdminError(username));
+        }
+
         this.getUserService().removeUser(username);
         Map<String, String> result = new HashMap<>();
         result.put("code", username);
         return new ResponseEntity<>(new RestResponse(result), HttpStatus.OK);
     }
 
+    private BindingResult createDeleteAdminError(@PathVariable String username) {
```
joewhite101: This needs test coverage for master

entando-service:
All is well

engine/src/main/java/org/entando/entando/web/user/UserController.java
rzsouza: Added missing unit tests and made some small improvements to the code

entando-service:
All is well

PR #725, opened 2018-11-13
EN-2178 Small configuration changes to avoid conflict of Guava version

rzsouza:
I noticed that cms and swagger weren't working at the same time because of different guava versions. It should work fine with this change


entando-service:
All is well

PR #727, opened 2018-11-14
EN-2178 Fix for Swagger

rzsouza:
Fix for both Guava dependency clash with CMS and also to make Swagger work with Spring Security


entando-service:
All is well

PR #728, opened 2018-11-14
Added missing improvements

eugeniosant:
Fixed CustomLocalizedProvider - Added option for properties into war (Jboss/Wildfly)
EN-1840 fix label in widget info
EN-1799 Add test for validation of widget type code length


entando-service:
All is well

PR #735, opened 2018-11-23
EN-2259 Changed RestResponse metadata using generics

zonia3000:


entando-service:
All is well

entando-service:
All is well

PR #738, opened 2018-11-30
SUP-97 Changed naming of uploaded files to keep name…

rzsouza:
Changed naming of uploaded files to keep name of the file similar to original name


entando-service:
All is well

PR #740, opened 2018-12-05
EN-2273 Complete OAuth2 flow

eugeniosant:
Implemented refresh token flow


entando-service:

admin-console/src/test/java/com/agiletec/apsadmin/admin/TestBaseAdminAction.java
```diff
@@ -1,131 +1,134 @@
...
+        this.setUserOnSession("supervisorCoach");
+        this.initAction("/do/BaseAdmin", "reloadConfig");
+        String result = this.executeAction();
+        assertEquals("userNotAllowed", result);
+
+        this.setUserOnSession("admin");
+        this.initAction("/do/BaseAdmin", "reloadConfig");
+        result = this.executeAction();
+        assertEquals(Action.SUCCESS, result);
+        synchronized (this) {
```
joewhite101: Why is this wait needed?

engine/src/main/java/org/entando/entando/aps/system/services/oauth2/ApiOAuth2TokenManager.java
```diff
@@ -30,17 +28,30 @@
...
+    private transient ScheduledExecutorService scheduler = null;
 
     private IOAuth2TokenDAO oAuth2TokenDAO;
 
     @Override
     public void init() throws Exception {
+        ScheduledDeleteExpiredTokenThread sdett
+                = new ScheduledDeleteExpiredTokenThread(this.getOAuth2TokenDAO(), this.getRefreshTokenValiditySeconds());
+        this.scheduler = Executors.newScheduledThreadPool(1);
+        this.scheduler.scheduleAtFixedRate(sdett, 0, 1, TimeUnit.HOURS);
```
joewhite101: shouldn't this be configuarble?

engine/src/test/java/com/agiletec/aps/system/services/user/AuthenticationProviderManagerTest.java
```diff
@@ -198,55 +199,103 @@ public void authenticate_1() throws Exception {
...
+        Assert.assertEquals(2, userDetails.getAuthorities().size());
+        Mockito.verify(userManager, Mockito.times(1)).updateLastAccess(Mockito.any(UserDetails.class));
+        Mockito.verify(userManager, Mockito.times(0)).getUser(Mockito.anyString(), Mockito.anyString());
+        Mockito.verify(userManager, Mockito.times(1)).getUser(Mockito.anyString());
+        Mockito.verify(authorizationManager, Mockito.times(1)).getUserAuthorizations(Mockito.anyString());
+        Mockito.verifyZeroInteractions(this.tokenManager);
+    }
+
+    @Test(expected = UsernameNotFoundException.class)
+    public void loadUserByUsername_2() throws Exception {
```
joewhite101: Like test loadUserByUsernamNotFound

engine/src/test/java/com/agiletec/aps/system/services/user/AuthenticationProviderManagerTest.java
```diff
@@ -198,55 +199,103 @@ public void authenticate_1() throws Exception {
...
+                    = this.authenticationProviderManager.loadUserByUsername("username");
+            Assert.fail();
+        } catch (UsernameNotFoundException e) {
+            Mockito.verifyZeroInteractions(this.authorizationManager);
+            Mockito.verifyZeroInteractions(this.tokenManager);
+            throw e;
+        }
+    }
+
+    @Test(expected = RuntimeException.class)
```
joewhite101: Is RuntimeException the specific expectation here? Seems to generic. Should be asserting against the UsernameNotFoundException. Generic RuntimeException check could mask a real error (though unlikely).

entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/services/oauth2/OAuth2TokenDAO.java
```diff
@@ -218,5 +245,45 @@ public void deleteExpiredToken() {
...
+                UsernamePasswordAuthenticationToken auth = new UsernamePasswordAuthenticationToken(username, "");
+                String clientid = res.getString("clientid");
+                Map<String, String> requestParameters = new HashMap<>();
+                requestParameters.put(OAuth2Utils.GRANT_TYPE, res.getString("granttype"));
+                OAuth2Request oAuth2Request = new OAuth2Request(requestParameters, clientid, null, true, null, null, null, null, null);
+                authentication = new OAuth2Authentication(oAuth2Request, auth);
+            }
+        } catch (Exception t) {
+            logger.error("Error while reading tokens", t);
+            throw new RuntimeException("Error while reading tokens", t);
```
joewhite101: Can we throw something more specific here? Not sure if it is totally required or not but a generic "throw RuntimeException" isn't great if avoidable

engine/src/main/java/org/entando/entando/aps/system/services/oauth2/IOAuth2TokenDAO.java
```diff
@@ -30,7 +31,13 @@
     public OAuth2AccessToken getAccessToken(final String accessToken);
 
     public void deleteAccessToken(final String accessToken);
+    
+    public void deleteExpiredToken(int expirationTime);
 
-    public void deleteExpiredToken();
+    public OAuth2RefreshToken readRefreshToken(String tokenValue);
```
joewhite101: Why "read" instead of "get"

admin-console/src/test/java/com/agiletec/apsadmin/admin/TestBaseAdminAction.java
eugeniosant: As it happens in other tests, with the synchronized block I give the time necessary for the underlying thread to start. Sometimes this test fails when the thread (which sets the expected value) is not started yet

engine/src/test/java/com/agiletec/aps/system/services/user/AuthenticationProviderManagerTest.java
eugeniosant: I used a identical name to the name of the method to test (with a suffix for different cases), but i can change it

engine/src/main/java/org/entando/entando/aps/system/services/oauth2/OAuth2TokenDAO.java
eugeniosant: Correct, we can throw something specific for the context (like an SQLException or something like that) but we need to decide if change it only here or apply a general strategy in every dao in the platform (now, in ours DAOs, we are using RuntimeException without any "throws" in method declaration)

engine/src/main/java/org/entando/entando/aps/system/services/oauth2/IOAuth2TokenDAO.java
eugeniosant: This dao is related to the TokenStore (interface defined by Spring Security) and I leaved the same names where it's possible
https://github.com/spring-projects/spring-security-oauth/blob/4d5517c6597502d769e75bff23b587ee86f3e271/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/TokenStore.java#L44

engine/src/main/java/org/entando/entando/aps/system/services/oauth2/ApiOAuth2TokenManager.java
joewhite101: Reviewed with Eugenio. This is just a cleanup thread for old tokens

engine/src/test/java/com/agiletec/aps/system/services/user/AuthenticationProviderManagerTest.java
joewhite101: Ok, lets change it to describe the purpose somewhat

engine/src/main/java/org/entando/entando/aps/system/services/oauth2/OAuth2TokenDAO.java
joewhite101: We can update these as we encounter them or make changes. In general, we should prefer specific exceptions

engine/src/main/java/org/entando/entando/aps/system/services/oauth2/IOAuth2TokenDAO.java
joewhite101: Ok that is fine for me as long as it matches a known pattern

engine/src/test/java/com/agiletec/aps/system/services/user/AuthenticationProviderManagerTest.java
eugeniosant: Done

entando-service:
All is well

entando-service:
All is well

engine/src/test/java/com/agiletec/aps/system/services/user/AuthenticationProviderManagerTest.java
eugeniosant: Fixed

engine/src/main/java/org/entando/entando/aps/system/services/oauth2/OAuth2TokenDAO.java
joewhite101: Reviewed with Eugenio and we'll leave this for now. Throwing a more specific exception trickles all the way back to an implementation of a spring interface. We would need to rework to much of the infrastructure to update it.

PR #742, opened 2018-12-05
Fixed infinite loop in getRefreshToken method

rzsouza:
Fixed a call of the method getRefreshToken to itself to avoid stack trace blowing up if someone calls it.

I also tidyied it up a bit fixing field names and other minor things.


entando-service:
All is well

entando-service:
All is well

PR #746, opened 2018-12-07
Removed duplicated condition in endWithParentDir method and some tidy up

rzsouza:
Minor fix and tidy up


entando-service:
All is well

PR #747, opened 2018-12-11
EN-2265 Fixed DE component lastUpdate date format and other little improvements

zonia3000:


entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/services/digitalexchange/model/ResilientPagedMetadata.java
```diff
@@ -30,10 +30,12 @@
 
     public ResilientPagedMetadata() {
         super();
+        this.errors = new ArrayList<>();
```
joewhite101: Any reason not initialize this on the member variable?
```
@JsonIgnore
private List<RestError> errors = new ArrayList<>();
```

zonia3000:
Hi @joewhite101, I added the "Work in Progress" label because Giuseppe reported me an issue with components filtering. I'll fix it and I'll commit changes on this branch since it is related to EN-2265 task

engine/src/main/java/org/entando/entando/aps/system/services/digitalexchange/model/ResilientPagedMetadata.java
zonia3000: No, I moved it there.

entando-service:

entando-service:
All is well

zonia3000:
@joewhite101 As pointed out by Eugenio, the default operator should be "eq" for each filter, however inside [Filter class](https://github.com/entando/entando-core/blob/master/engine/src/main/java/org/entando/entando/web/common/model/Filter.java) the default operator is "like". I tried to change like into eq but tests failed. I think that we should try to align the default operator in the various parts of the code, but probably this requires a dedicated ticket.

entando-service:
All is well

PR #748, opened 2018-12-11
Updated some deprecated classes

rzsouza:
And minor tidy up

### DIFF
--- a/engine/src/main/java/org/entando/entando/aps/system/services/page/PageTokenManager.java
+++ b/engine/src/main/java/org/entando/entando/aps/system/services/page/PageTokenManager.java
@@ -100,7 +100,7 @@ public class PageTokenManager extends AbstractService implements IPageTokenManag
 			pbeCipher.init(Cipher.DECRYPT_MODE, key, new PBEParameterSpec(this.getSalt().getBytes(), 20));
 			return new String(pbeCipher.doFinal(base64Decode(property)), "UTF-8");
 		} catch (GeneralSecurityException | IOException e) {
-			logger.error("Error in decrypt");
+			logger.error("Error in decrypt", e);
 		}
 		return null;
 	}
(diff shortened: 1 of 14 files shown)


rzsouza:
I just saw that some of the files had their EOL characters fixed during the upload. They all had minor changes and it's probably easier to check them in the side by side mode

entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/services/page/PageTokenManager.java
```diff
@@ -102,11 +99,9 @@ public String decrypt(String property) {
 			Cipher pbeCipher = Cipher.getInstance("PBEWithMD5AndDES");
 			pbeCipher.init(Cipher.DECRYPT_MODE, key, new PBEParameterSpec(this.getSalt().getBytes(), 20));
 			return new String(pbeCipher.doFinal(base64Decode(property)), "UTF-8");
-		} catch (GeneralSecurityException e) {
-			logger.error("Error in decrypt");
-		} catch (IOException e) {
+		} catch (GeneralSecurityException | IOException e) {
 			logger.error("Error in decrypt");
```
joewhite101: If we're changing this we should also log the exception itself `logger.error("Error in decrypt",e);

engine/src/main/java/org/entando/entando/aps/system/services/storage/api/LocalStorageManagerInterface.java
```diff
@@ -40,10 +39,15 @@
 
 public class LocalStorageManagerInterface implements IApiExportable {
 
-	private static final Logger _logger = LoggerFactory.getLogger(LocalStorageManagerInterface.class);
+	private static final Logger logger = LoggerFactory.getLogger(LocalStorageManagerInterface.class);
+
+	private IStorageManager _storageManager;
```
joewhite101: If touching remove underscore

engine/src/main/java/org/entando/entando/aps/system/services/page/PageTokenManager.java
rzsouza: Fixed

engine/src/main/java/org/entando/entando/aps/system/services/storage/api/LocalStorageManagerInterface.java
rzsouza: Fixed

entando-service:
All is well

PR #749, opened 2018-12-13
Fixed filtering that wasn't working properly

rzsouza:
While I was testing for EN-2162, I found a filter that wasn't working. I added a test and fixed it.

### DIFF
--- a/engine/src/main/java/org/entando/entando/aps/system/services/entity/AbstractEntityTypeService.java
+++ b/engine/src/main/java/org/entando/entando/aps/system/services/entity/AbstractEntityTypeService.java
@@ -13,67 +13,42 @@
  */
 package org.entando.entando.aps.system.services.entity;
 
-import java.lang.reflect.InvocationTargetException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.stream.Collectors;
-
-import com.agiletec.aps.system.common.IManager;
-import com.agiletec.aps.system.common.entity.IEntityManager;
-import com.agiletec.aps.system.common.entity.IEntityTypesConfigurer;
-import com.agiletec.aps.system.common.entity.model.ApsEntity;
-import com.agiletec.aps.system.common.entity.model.IApsEntity;
-import com.agiletec.aps.system.common.entity.model.attribute.AbstractListAttribute;
-import com.agiletec.aps.system.common.entity.model.attribute.AttributeInterface;
-import com.agiletec.aps.system.common.entity.model.attribute.CompositeAttribute;
-import com.agiletec.aps.system.common.entity.model.attribute.EnumeratorAttribute;
-import com.agiletec.aps.system.common.entity.model.attribute.util.EnumeratorAttributeItemsExtractor;
-import com.agiletec.aps.system.common.entity.model.attribute.util.IAttributeValidationRules;
+import com.agiletec.aps.system.common.entity.*;
+import com.agiletec.aps.system.common.entity.model.*;
+import com.agiletec.aps.system.common.entity.model.attribute.*;
+import com.agiletec.aps.system.common.entity.model.attribute.util.*;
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/entando/entando/aps/system/services/entity/AbstractEntityTypeService.java
```diff
@@ -100,15 +75,25 @@ protected EntityTypeShortDto toDto(IApsEntity src) {
...
 
-    protected PagedMetadata<EntityTypeShortDto> getShortEntityTypes(String entityManagerCode, RestListRequest requestList) {
+    protected PagedMetadata<EntityTypeShortDto> getShortEntityTypes(String entityManagerCode,
+                                                                    RestListRequest requestList) {
+
         IEntityManager entityManager = this.extractEntityManager(entityManagerCode);
-        List<IApsEntity> entityTypes = new ArrayList<>(entityManager.getEntityPrototypes().values());
+
         Map<String, String> fieldMapping = this.getEntityTypeFieldNameMapping();
-        entityTypes.stream().filter(i -> this.filterObjects(i, requestList.getFilters(), fieldMapping));
```
rzsouza: This filter here doesn't work because the streams are lazy. It would only work if we added a collect and retrieved the result in a new variable.

engine/src/main/java/org/entando/entando/aps/system/services/entity/AbstractEntityTypeService.java
```diff
@@ -100,15 +75,25 @@ protected EntityTypeShortDto toDto(IApsEntity src) {
...
         if (!RestListRequest.DIRECTION_VALUE_DEFAULT.equals(requestList.getDirection())) {
-            Collections.reverse(entityTypes);
+            caseInsensitiveComparator.reversed();
         }
+
+        List<IApsEntity> entityTypes = entityManager.getEntityPrototypes()
+                .values()
+                .stream()
+                .filter(i -> this.filterObjects(i, requestList.getFilters(), fieldMapping))
+                .sorted(new BeanComparator<>(getFieldName(requestList.getSort(), fieldMapping), caseInsensitiveComparator))
```
rzsouza: @zonia3000 Here I'd rather have used your CaseInsensitiveComparator, but Java was giving me some grief because of the types of your comparator, I replaced with a TransformerComparator in place. Maybe you have a better idea about how to use your comparator in the stream here

entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/services/entity/AbstractEntityTypeService.java
zonia3000: Hi Rodrigo, `CaseInsensitiveBeanComparator` can't be very typed, so I think that your solution is better. We should remove that class and its test if it is not used anymore.

engine/src/main/java/org/entando/entando/aps/system/services/entity/AbstractEntityTypeService.java
```diff
@@ -360,14 +356,13 @@ protected IEntityManager extractEntityManager(String entityManagerCode) {
...
             }
             Map<String, String> properties = BeanUtils.describe(bean);
             for (Filter filter : filters) {
                 String fieldName = this.getFieldName(filter.getAttribute(), mapping);
-                String value = (null != properties.get(fieldName))
-                        ? properties.get(fieldName) : null;
+                String value = properties.get(fieldName);
                 if (null == value) {
                     continue;
                 }
```
zonia3000: Here the filtering is ignoring the operator.
I fixed the same issues for other classes in this PR, that is still open:
https://github.com/entando/entando-core/pull/747
I can fix this in my branch after this PR has been merged.

engine/src/main/java/org/entando/entando/aps/system/services/entity/AbstractEntityTypeService.java
rzsouza: In this case, I'll remove that class and move the new comparator code into a separate class to make it easier to be reused.

PR #752, opened 2018-12-17
EN-2277 Changed annotation check to 

rzsouza:
consider RequestMapping annotation aliases such as GetMapping, PostMapping, PutMapping...


entando-service:
All is well

PR #756, opened 2018-12-19
Minor tidy up and test fix

rzsouza:
We had some conditions in the format 
`name.trim().length()>=0`

These are always true. I removed them and tidied up the files that I touched.

I also fixed one test that looked like it was running but it wasn't really testing anything, just failing in silence

### DIFF
--- a/admin-console/src/test/java/com/agiletec/apsadmin/system/entity/TestEntityManagersAction.java
+++ b/admin-console/src/test/java/com/agiletec/apsadmin/system/entity/TestEntityManagersAction.java
@@ -31,9 +31,6 @@ import com.agiletec.apsadmin.system.entity.type.IEntityTypeConfigAction;
 import com.agiletec.apsadmin.system.entity.type.IEntityTypesAction;
 import com.opensymphony.xwork2.Action;
 
-/**
- * @author E.Santoboni
- */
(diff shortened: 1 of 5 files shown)


admin-console/src/test/java/com/agiletec/apsadmin/system/entity/TestEntityManagersAction.java
```diff
@@ -230,30 +222,25 @@ public void testSaveNewAttribute() throws Throwable  {
...
-			//duplicate attribute name
-			attributeName="right_name";
-			result = this.executeSaveAttribute("admin", attributeName, attributeTypeCode, entityTypeCode, strutsAction, null, null, entityManagerName);
-			assertEquals(Action.INPUT, result);
+			action = (EntityAttributeConfigAction) getAction();
 			fieldErrors = action.getFieldErrors();
-			assertEquals(1, fieldErrors.size());
-			assertTrue(fieldErrors.containsKey("attributeName"));
+			assertTrue(fieldErrors.isEmpty());
 		} catch (Throwable t) {
```
rzsouza: Because of this catch in Throwable, and the empty catch block, this test was raising an error in one of the first asserts and after that it was aborting because the catch was swallowing the assert error.

entando-service:
All is well

PR #757, opened 2018-12-21
EN-2289 EN-2265 DE API changes

zonia3000:


zonia3000:
The main change on this PR involves the `combineResults` method of the `DigitalExchangeCall` class. Its parameter has been changed from `List<PagedRestResponse<T>> allResults` to `Map<String, PagedRestResponse<T>> allResults`: in this way it is possible to know which DE instance generated each result. This is useful both for the components endpoint (so that for each component we can know the name of the DE providing it) and for the Test All endpoint (so that we know the DE instance that generated a specific error).

**Important**: after this PR has been merged the digital-exchange project will not compile until [this branch](https://github.com/entando/digital-exchange/tree/EN-2289) will be merged.

entando-service:
All is well

PR #739, opened 2018-12-03
EN-2265 Created API for retrieving components from Digital Exchange instances

zonia3000:

PR #759, opened 2019-01-07
EN-2333 Removed DE code from core

zonia3000:


zonia3000:
I was able to move everything except some of the messages contained in `rest/messages.properties`, because the `MessageSource` bean is created in the engine and I can't find a way for retrieving the messages from outside.
I had a look at [ReloadableResourceBundleMessageSource](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/support/ReloadableResourceBundleMessageSource.html), but I don't know how it could be configured in a situation in which one have multiple plugins as dependencies and each one provides its own messages.

In any case, I moved back the `messageSource` bean definition into `servlet-context.xml` (I've previously moved it into `servicesConfig.xml` due to DE client needs, see #739) and I defined a new message source bean in DE Plugin. So, at least messages used by DE client are contained in DE plugin.

Related branch in components is [here](https://github.com/entando/entando-components/tree/EN-2333).

entando-service:
All is well

rzsouza:
All good

PR #764, opened 2019-01-17
EN-2312 Changes for supporting installation of a single component

zonia3000:
Just made some methods public, introduced two interfaces and a bit of refactoring.

### DIFF
--- a/engine/src/main/java/org/entando/entando/aps/system/init/InitializerManager.java
+++ b/engine/src/main/java/org/entando/entando/aps/system/init/InitializerManager.java
@@ -180,7 +180,7 @@ public class InitializerManager extends AbstractInitializerManager implements II
 
 	//-------------------- REPORT -------- START
 
-	private void saveReport(SystemInstallationReport report) throws BeansException {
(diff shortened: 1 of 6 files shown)


engine/src/main/java/org/entando/entando/aps/system/init/InitializerManager.java
```diff
@@ -37,164 +37,164 @@
...
+        logger.debug("{}: initializated - Check on startup {}", this.getClass().getName(), this.isCheckOnStartup());
+    }
+
+    public void executePostInitProcesses() throws BeansException {
+        SystemInstallationReport report = null;
+        try {
+            report = this.extractReport();
+            List<Component> components = this.getComponentManager().getCurrentComponents();
+            for (Component component : components) {
+                executeComponentPostInitProcesses(component, report);
```
rzsouza: Best improvement of this changeset, very good

zonia3000: thanks!

entando-service:
All is well

PR #765, opened 2019-01-18
EN - 2338 JSON patch applied to JSON object 

Kerruba:
With this pull request I'm supporting JSON patch operations on the Page model. 

This implementation works by converting the POJO to a JSON object, applying the patch to the JSON object and converting the JSON object back to a POJO.

Also, the update to the object is applied using the update method provided for the PUT request, so it's reusing that functionality


entando-service:
All is well

Kerruba:
We have decided to use the Spring methodology to deal with Patches

PR #766, opened 2019-01-18
EN-2349 Fixed PageService issue and added tests

zonia3000:
Now, during an update, the `PageService` removes all the extra groups before adding the ones coming from the request, so that groups removal is working properly on update. I added unit test for this case.

Moreover I checked the `movePage()` method because I remembered the issues described in EN-2159. The various checks necessary for the validation of the page movement were ok, however the logic for the group checking was partially replicated in `PageService` and in `PageValidator`. I moved it on the validator and then I added the test for these cases (the check of the invalid "free page" under "reserved parent" was not tested according to Sonar).


entando-service:
All is well

PR #767, opened 2019-01-18
EN-2338 JSON patch using Spring Patch

Kerruba:
This is another implementation of the JSON patch, but this time using Spring Patch implementation. 

I think this implementation has some limitations as already discussed, and I would prefer the other approach even if not officially supported by Spring. (Here the [other pull request](https://github.com/entando/entando-core/pull/765) with my favorite approach)

Maybe we can park this implementation until Spring Patch will be able to support updates to Maps 

WIP, do not merge yet


entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/services/page/PageService.java
```diff
@@ -278,6 +287,29 @@ public PageDto updatePage(String pageCode, PageRequest pageRequest) {
...
+
+
+
+        try {
+            PageDto pageDto = this.getDtoBuilder().convert(oldPage);
+            PageDto updatedPageDto = this.getJsonPatchService().applyPatch(jsonPatch, pageDto);
+            PageRequest updateRequest = new PageDtoToPageRequestConverter().convert(updatedPageDto);
+            return this.updatePage(pageCode, updateRequest);
+        } catch (Exception e) {
+            e.printStackTrace();
```
zonia3000: Use a logger here

engine/src/main/java/org/entando/entando/web/page/model/PageRequest.java
```diff
@@ -17,27 +17,29 @@
...
+//import org.hibernate.validator.constraints.NotBlank;
 
 /**
  *
  * @author paddeo
  */
 public class PageRequest {
 
-    @NotBlank(message = "page.code.NotBlank")
+    @NotNull(message = "page.code.NotBlank")
```
zonia3000: I prefer using the javax.validation classes too, but I think here we need also a `@Size` constraint.

engine/src/main/java/org/entando/entando/web/page/PageController.java
```diff
@@ -35,19 +44,20 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.data.rest.webmvc.RestMediaTypes;
```
zonia3000: Unused import

engine/src/main/java/org/entando/entando/web/page/validator/PageValidator.java
```diff
@@ -109,4 +123,21 @@ public void validatePagesStatus(String pageCode, PagePositionRequest pageRequest
...
+            jsonPatchValidator.validatePatch(jsonPatch);
+        } catch (InvalidJsonPatchException e) {
+            errors.reject(PageController.ERRCODE_INVALID_PATCH, new String[]{pageCode}, "page.update.patch.invalid");
+        }
+
+        for (JsonNode node : jsonPatch) {
+            String operationPath = node.get("path").asText();
+
+            if (operationPath.equals("/code")) {
+                errors.reject(PageController.ERRCODE_INVALID_PATCH, new String[]{pageCode}, "page.update.patch.invalid.code");
```
zonia3000: Here you can pass an empty array to the reject method, because the pageCode is not used in the `page.update.patch.invalid.code` error message.

engine/src/main/java/org/entando/entando/aps/system/services/page/PageService.java
Kerruba: Should I throw the error anyway right?

engine/src/main/java/org/entando/entando/web/page/model/PageRequest.java
zonia3000: I think that we have to keep the same constraints that are actually used in the action. If you look at the `PageAction-validation.xml` (inside admin-console module) you can see that the size can be between 1 and 30 and there is also a regex validation.

engine/src/main/java/org/entando/entando/aps/system/services/page/PageService.java
zonia3000: Maybe you can catch a more specific exception (like the `PatchException` that can be thrown by the `applyPatch()` method) and rethrow a `RestServerError`, like the other methods are doing.

entando-service:
All is well

entando-service:
All is well

entando-service:
All is well

entando-service:
All is well

entando-service:

entando-service:
All is well

entando-service:
All is well

entando-service:

entando-service:

zonia3000:
retest this please

entando-service:

entando-service:
All is well

entando-service:

entando-service:
All is well

PR #772, opened 2019-02-04
En 2311 components uninstall

Kerruba:
This PR is to start supporting DE components uninstall

### DIFF
--- a/engine/src/main/java/org/entando/entando/aps/system/init/DatabaseManager.java
+++ b/engine/src/main/java/org/entando/entando/aps/system/init/DatabaseManager.java
@@ -433,10 +433,8 @@ public class DatabaseManager extends AbstractInitializerManager
                     }
                 }
 
-
-                report.setStatus(SystemInstallationReport.Status.UNINSTALLED);
+                report.removeComponentReport(componentConfiguration.getCode());
                 report.setUpdated();
-
                 System.out.println(logPrefix + "\n" + logPrefix + "Uninstall complete\n" + logPrefix);
                 logger.debug(logPrefix + "\n" + logPrefix + "Uninstall complete\n" + logPrefix);
             } catch (Throwable t) {
--- a/engine/src/main/java/org/entando/entando/aps/system/init/DatabaseManager.java
+++ b/engine/src/main/java/org/entando/entando/aps/system/init/DatabaseManager.java
@@ -25,10 +25,15 @@ import java.util.Map;
 import javax.servlet.ServletContext;
 import javax.sql.DataSource;
 
+import com.agiletec.aps.system.exception.ApsSystemException;
+import com.agiletec.aps.util.ApsWebApplicationUtils;
+import com.agiletec.aps.util.DateConverter;
+import com.agiletec.aps.util.FileTextReader;
 import org.apache.commons.beanutils.BeanComparator;
 import org.entando.entando.aps.system.init.model.Component;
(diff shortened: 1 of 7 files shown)


entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/init/DatabaseManager.java
```diff
@@ -391,6 +391,59 @@ public void initComponentDefaultResources(Component componentConfiguration,
         }
     }
 
+    public void uninstallComponentResources(Component componentConfiguration,
+                                            SystemInstallationReport report) throws ApsSystemException {
+        String logPrefix = "|   ";
```
zonia3000: A general improvement on this class: we could move this string in a static field, since it is used in various method.

engine/src/main/java/org/entando/entando/aps/system/init/DatabaseManager.java
```diff
@@ -391,6 +391,59 @@ public void initComponentDefaultResources(Component componentConfiguration,
...
+        System.out.println("+ [ Component: " + componentConfiguration.getCode() + " ] :: DATA\n" + logPrefix);
+        ComponentInstallationReport componentReport = report.getComponentReport(componentConfiguration.getCode(), false);
+        if (componentReport.getStatus().equals(SystemInstallationReport.Status.UNINSTALLED)) {
+
+            logger.debug(logPrefix + "( ok ) Already uninstalled\n" + logPrefix);
+            System.out.println(logPrefix + "( ok ) Already uninstalled\n" + logPrefix);
+            return;
```
zonia3000: This return is not necessary

entando-service:
All is well

engine/src/test/java/org/entando/entando/aps/system/init/DatabaseManagerTest.java
```diff
@@ -0,0 +1,75 @@
...
+    private Component component;
+
+    @Mock
+    private IInitializerManagerCacheWrapper cacheWrapper;
+
+    @Mock
+    private ListableBeanFactory beanFactory;
+
+    @InjectMocks
+    private DatabaseManager databaseManager = null;
```
zonia3000: This `= null` is not necessary

engine/src/test/java/org/entando/entando/aps/system/init/DatabaseManagerTest.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Test
+    public void should_uninstall_component() throws ApsSystemException, IOException {
+        component = ComponentUtils.getEntandoComponent("test_component");
+
+        when(beanFactory.getBeanNamesForType(DataSource.class)).thenReturn(new String[]{"portDataSource", "servDataSource"});
+
+        SystemInstallationReport systemInstallationReport = createMockReport();
+        databaseManager.uninstallComponentResources(component, systemInstallationReport);
```
zonia3000: If you want you can wrap the fake report using the Mockito `spy()` method. Then you can use the `verify()` method on the object for checking specific calls performed by the tested method, like this:

```java
SystemInstallationReport systemInstallationReport = spy(createMockReport());
databaseManager.uninstallComponentResources(component, systemInstallationReport);
verify(systemInstallationReport, times(1)).removeComponentReport("test_component");
```

entando-service:
All is well

PR #774, opened 2019-02-12
EN-2378 Filter improvements

zonia3000:


entando-service:
All is well

engine/src/main/java/org/entando/entando/web/common/model/Filter.java
```diff
@@ -90,6 +93,13 @@ public EntitySearchFilter getEntitySearchFilter() {
         return filter;
     }
 
+    public List<String> getFilterValues() {
```
Kerruba: Since this is basically made for the OR conditions, to make it more explicit and easier to interpret without the need for comments what if we change the name to something like "getOrConditions" or "getOrValues" or something like that?

engine/src/main/java/org/entando/entando/aps/util/FilterUtils.java
```diff
@@ -78,33 +102,53 @@ public static boolean filterLong(Filter filter, Supplier<Long> supplier) {
...
+                    break;
+                default:
+                    throw new UnsupportedOperationException(getUnsupportedOperatorMessage(filter));
+            }
         }
+
+        return result;
+    }
+
+    private static <T> List<T> getTypedValues(Filter filter, Function<String, T> converter) {
```
Kerruba: Check the next comment on a suggestion on clearer naming

entando-service:
All is well

zonia3000:
Please, before merging be sure that [PR #486 in components](https://github.com/entando/entando-components/pull/468) has been approved.

PR #780, opened 2019-02-22
Solves a minor issue with css of the AdminUI

Kerruba:
Adds necessary div and class to not break admin ui for Date attribute validation


entando-service:
All is well

PR #784, opened 2019-03-04
EN-2405 Improvement of single rest filter (5.0.3)

eugeniosant:


entando-service:
All is well

eugeniosant:
This PR fix a little bug, in the current version (5.0.x) it's not possible to create a filter with clauses "lower than" and "greater than". In master this bug is already fixed. My target was to create a simple patch to apply (when we need) in projects on version 5.0.2

PR #789, opened 2019-03-11
En 2413 fix duplicate key on tokens

ampie:
I encountered an intermittent duplicate key exception on storing the OAuth Token. This only happens on MySQL and only on certain nodes.  The code in the DAO that inserts the token checks for the existience of the token just before it inserts it. If the token already exists, the method returns immediately. The implication of this exception is that this check does not detect the token yet, but it gets inserted exactly as the rest of the method is executed. As a fix, I repeated this check in the 'catch' clause if the exception involved a duplicate key contraint violation, and then swallow the exception if the token already exists. This got rid of the intermittent exception entirely,


entando-service:

entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/services/oauth2/OAuth2TokenDAO.java
```diff
@@ -190,6 +193,18 @@ public void storeAccessToken(OAuth2AccessToken accessToken, OAuth2Authentication
             stat.executeUpdate();
             conn.commit();
         } catch (Exception t) {
+            CharArrayWriter stackTraceWriter = new CharArrayWriter();
```
joewhite101: I don't think we can merge this. Looks like a potential security problem. We'll have to find the underlying race condition.

joewhite101:
Overcome by events.

PR #793, opened 2019-03-26
EN-2368 Fixing label filtering and adding test coverage.

sergiofilhowz:
Fixing label filtering and adding test coverage.

P.S: I did not use the sort property because the old code didn't use as well, so I imagine it might break things if I do.


entando-service:
All is well

entando-service:
All is well

PR #795, opened 2019-03-27
EN-2427 EN-2423 Updated bootstrap-switch again

zonia3000:
It seems that there was something wrong with the min.js file. I simply downloaded it again from GitHub and overwrote it. This file was recently updated with PR #791 in order to fix an IE 11 bug. I would double check the behavior with IE 11.


entando-service:
All is well

PR #805, opened 2019-05-09
EN-2393 Fixed issues in Consumer API

zonia3000:
It seems that there is a timezone conversion issue when Jackson deserialize the dates (I found some threads on GitHub about this). To solve the issue I changed the field type from `Date` to `String` and I performed the date validation/conversion manually (using the `DateConverter` class already provided by the Entando core).


entando-service:
All is well

PR #812, opened 2019-07-10
Updates mockito version

Kerruba:
This fixes issue with entando-de-app not running with jetty-container profile


entando-service:
All is well

PR #816, opened 2019-09-11
EN6-97 i18n for the CMS endpoints

ffleandro:
i18n for endpoints for Content Settings, Digital Assets and Contents;


entando-service:
All is well

entando-service:
All is well

PR #817, opened 2019-09-18
Update API rest server to pass current user through to child methods …

joewhite101:
…for PAM


entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/services/api/server/ApiRestServer.java
```diff
@@ -283,6 +283,7 @@ else if (token.isExpired()) {
                 String username = token.getLocalUser();
                 IUserManager userManager = (IUserManager) ApsWebApplicationUtils.getBean(SystemConstants.USER_MANAGER, request);
                 UserDetails user = userManager.getUser(username);
+                properties.put(SystemConstants.API_USER_PARAMETER, user);
```
eugeniosant: The user can be null and in Properies neither the key or the value can be null. 
This line is better after check of user

entando-service:
All is well

PR #818, opened 2019-10-01
EN6-103 separate user session extraction from permission validation

ffleandro:
In the opened ticket I noticed that some endpoints didn't have the `@RestAccessControl` annotation but still had User dependent behaviour.
However the logged in user was never being processed because of the lack of the annotation.

I changed the `EntandoOauth2Interceptor` to separate the business logic between the methods `extractOAuthParameters` and `checkPermissions`, as I noticed they were too entangled.

`extractOAuthParameters` now doesn't make any validations or throws any errors, it only extracts the `accessToken ` and loads the corresponding user if available.

`checkPermissions` now is only checked if there is a `@RestAccessControl` configured with a `permission`.
If not configured, the `UserDetails` is still being loaded into the `HttpSession`.

### DIFF
--- a/engine/src/main/java/org/entando/entando/web/common/interceptor/EntandoOauth2Interceptor.java
+++ b/engine/src/main/java/org/entando/entando/web/common/interceptor/EntandoOauth2Interceptor.java
@@ -58,7 +58,7 @@ public class EntandoOauth2Interceptor extends HandlerInterceptorAdapter {
             if (method.hasMethodAnnotation(RequestMapping.class)) {
                 UserDetails user = this.extractOAuthParameters(request);
                 RestAccessControl rqm = method.getMethodAnnotation(RestAccessControl.class);
-                if (null == rqm || null == user) {
+                if (null == rqm) {
                     return true;
                 }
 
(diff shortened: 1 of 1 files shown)


entando-service:
All is well

engine/src/main/java/org/entando/entando/web/common/interceptor/EntandoOauth2Interceptor.java
```diff
@@ -56,54 +56,62 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
...
-        if (user != null) {
-            request.getSession().setAttribute("user", user);
-            logger.debug("User {} requesting resource that requires {} permission ", username, permission);
-            if (StringUtils.isNotBlank(permission)) {
-                if (!this.getAuthorizationManager().isAuthOnPermission(user, permission)) {
-                    logger.warn("User {} is missing the required permission {}", username, permission);
-                    throw new EntandoAuthorizationException(null, request, username);
-                }
+    protected void checkAuthorization(UserDetails user, String permission, HttpServletRequest request) throws ApsSystemException {
+        if (null == user) {
```
joewhite101: This is a pretty big change (even if an improvement). Is there a reason to change this flow to throw an exception when no user is present on the request? The original logged info and carried on.

ffleandro: This exception was already being thrown in the `extractOAuthParameters` method.

As I explained in the PR text, the main change on this PR is to separate responsabilities as both `extractOAuthParameters` and `checkPermission` were throwing exceptions.

So by separating both methods, one's responsability is to just extract the token and user if present and the second is to actually validate it only if the annotation is configured to require a permission (and hence needs an authenticated user).

Check the method `preHandle`, as you can see, first the user is extracted and only if the annotation/permission is configured that we call the `checkPermission` method to guarantee that there is a logged in user and it has the correct permissions.

PR #825, opened 2019-10-29
Bump version to 6.0.0-SNAPSHOT

Kerruba:
Next I'll update also the components (at least the digital-exchange and CMS) to this new version to support the entando-de-app


entando-service:
All is well

PR #835, opened 2019-12-02
EN6-178 Fixing lastPage values when totalItems are 0

BrenoQVDS:
Adding fix for lastPage value when totalItems are 0


entando-service:
All is well

PR #836, opened 2019-12-06
EN6-171 login screen of admin console for mobile / tablet

jeffgo10:
Login page now designed for mobile/tablet size


entando-service:
All is well

PR #838, opened 2019-12-11
EN6-143 Changing attributes values to match new requirements

BrenoQVDS:
Due to new requirements the values for searchableOptionSupported and indexableOptionSupported needed to me changed


entando-service:
All is well

PR #840, opened 2019-12-12
Reimplements the XSSRequestWrapper to validate the parameters and the…

Kerruba:
… headers and encode them to HTML attributes

Wait for feedback from @entsimsan before merging


entando-service:

entando-service:
All is well

PR #863, opened 2020-03-23
Fix bad count ref in page manager cache wrapper that prevents saving …

joewhite101:
…changes to pages with no widgets


entando-service:
All is well

PR #866, opened 2020-03-30
ENG-303 Add page, remove ir and add it again test

BrenoQVDS:
This is just a test to look for the problem, the API is looking fine


entando-service:
All is well

PR #874, opened 2020-04-09
ENG-331 Component usage endpoints

ffleandro:


entando-service:
All is well

engine/src/main/db/update/scriptServ.sql
```diff
@@ -1 +1,4 @@
-
+ALTER TABLE authusers ALTER COLUMN username TYPE character varying(80);
```
Kerruba, line 1: Was this intended? I can see these line were addressing a different sql script before (port) and now are in serv

joewhite101: Doesn't look right to me either? Is this the right file for this.

engine/pom.xml
```diff
@@ -258,6 +258,11 @@
         </pluginManagement>
     </build>
     <dependencies>
+        <dependency>
```
joewhite101, line 261: Can't do this. Will break something in EAP or Wildfly.

ffleandro:
This was intended. Those tables that were changed belonged to a different Entando database and weren’t running when I tried to upgrade my environment. 

Already checked with Breno whom authored those PRs.

joewhite101:
Ok the SQL query is ok based on the info below from Breno. Lombok still an issue though

We moved the user to a different database?/
1:50
For which story?

Breno Queiroz  1:54 PM
No, the problem was in the initial ticket I made, which was to allow users with longer characters. I put the update SQL in the Port database, when it was supposed to be in the Serv database.
Filipe noticed that and moved to the correct SQL

Joe White:rocket:  1:55 PM
ok so the column hasn’t moved?
New

Breno Queiroz  1:55 PM
No, I just updated the length

ffleandro:
Removed Lombok dependency and implemented the Builder pattern manually on ComponentUsage.

@joewhite101 and @Kerruba can you guys finish the review?

entando-service:
All is well

entando-service:

entando-service:
All is well

PR #882, opened 2020-04-27
ENG-437: Removed exception log if request body is empty

alepintus:
Hey @joewhite101 @Kerruba ,
this PR remove the logging of the exception if one widget without config is drag and dropped in the page configuration.

Could you please have a look?
thanks

### DIFF
--- a/engine/src/main/java/org/entando/entando/web/common/interceptor/ActivityStreamInterceptor.java
+++ b/engine/src/main/java/org/entando/entando/web/common/interceptor/ActivityStreamInterceptor.java
@@ -13,9 +13,6 @@
  */
 package org.entando.entando.web.common.interceptor;
 
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-
 import com.agiletec.aps.system.SystemConstants;
 import com.agiletec.aps.system.services.user.UserDetails;
 import com.agiletec.aps.util.FileTextReader;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/entando/entando/web/common/interceptor/ActivityStreamInterceptor.java
```diff
@@ -119,10 +118,15 @@ private String addParamsFromRequestBody(String prev, InputStream is) {
         try {
             CloseShieldInputStream cloned = new CloseShieldInputStream(is);
             String body = FileTextReader.getText(cloned);
-            byte[] bytes = body.getBytes();
-            ObjectMapper objectMapper = new ObjectMapper();
-            Map<String, String> params = objectMapper.readValue(bytes, HashMap.class);
-            return this.addParameters(prev, params);
+            int bodyLength = body.length();
```
Kerruba, line 121: Could it be that the body is null?

entando-service:
All is well

engine/src/main/java/org/entando/entando/web/common/interceptor/ActivityStreamInterceptor.java
alepintus: Hey @Kerruba ,
thanks for the review.
I did a code check and the variable body can't be null here.

PR #885, opened 2020-05-05
ENG-250: parent page update fixed in page edit

alepintus:
Hey @joewhite101 @Kerruba @eugeniosant ,
With this PR the user now can change the position of a page  (updating the parent page) also in the "edit page" page.
 
Could you please have a look?

Thanks

### DIFF
--- a/engine/src/main/java/org/entando/entando/aps/system/services/page/PageService.java
+++ b/engine/src/main/java/org/entando/entando/aps/system/services/page/PageService.java
@@ -13,15 +13,6 @@
  */
 package org.entando.entando.aps.system.services.page;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-
 import com.agiletec.aps.system.common.FieldSearchFilter;
 import com.agiletec.aps.system.common.IManager;
 import com.agiletec.aps.system.common.model.dao.SearcherDaoPaginatedResult;
@@ -41,6 +32,14 @@ import com.agiletec.aps.system.services.pagemodel.PageModel;
 import com.agiletec.aps.system.services.pagemodel.PageModelUtilizer;
 import com.agiletec.aps.util.ApsProperties;
 import com.fasterxml.jackson.databind.JsonNode;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import org.apache.commons.beanutils.BeanComparator;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/entando/entando/web/page/PageController.java
```diff
@@ -189,15 +187,42 @@ public void setAuthorizationService(PageAuthorizationService authorizationServic
...
+        pagePositionRequest.setPosition(position);
+
+        validateMovePage(pageCode, bindingResult, pagePositionRequest);
+
         PageDto page = this.getPageService().updatePage(pageCode, pageRequest);
         Map<String, String> metadata = new HashMap<>();
         return new ResponseEntity<>(new RestResponse<>(page, metadata), HttpStatus.OK);
     }
 
+    private void validateMovePage(@PathVariable String pageCode,
```
joewhite101, line 203: Is this the right place for this? Where does this same logic occur when we move a page on the main page tree? Can we call that instead?

engine/src/main/java/org/entando/entando/web/page/PageController.java
```diff
@@ -189,15 +187,42 @@ public void setAuthorizationService(PageAuthorizationService authorizationServic
...
+            BindingResult bindingResult, PagePositionRequest pagePositionRequest) {
+        this.getPageValidator().validateChangePositionRequest(pageCode, pagePositionRequest, bindingResult);
+
+        if (bindingResult.hasErrors()) {
```
joewhite101, line 207: Need to make sure these error conditions are covered by test cases

engine/src/main/java/org/entando/entando/aps/system/services/page/PageService.java
```diff
@@ -260,19 +259,24 @@ public PageDto updatePage(String pageCode, PageRequest pageRequest) {
...
-        }
+
         if (!oldPage.getGroup().equals(pageRequest.getOwnerGroup())) {
             BeanPropertyBindingResult bindingResult = new BeanPropertyBindingResult(oldPage, "page");
             bindingResult.reject(PageValidator.ERRCODE_GROUP_MISMATCH,
                     new String[]{oldPage.getGroup(), pageRequest.getOwnerGroup()}, "page.update.group.invalid");
             throw new ValidationGenericException(bindingResult);
         }
         try {
+            if (!oldPage.getParentCode().equals(pageRequest.getParentCode())) {
```
joewhite101, line 270: This logic has to exist elsewhere as well for the page move on the page tree. Can we combine or call it once?

engine/src/main/java/org/entando/entando/web/page/PageController.java
alepintus: This is a small refactoring, It's the old main page tree code moved to this new method (to avoid code duplication) and reused here.

entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/services/page/PageService.java
alepintus: I searched but I have not found any other check like this in the code. I added this check here to be sure that the user is moving the page from is previous parent branch to a new one.

engine/src/main/java/org/entando/entando/web/page/PageController.java
alepintus: New tests added

entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/services/page/PageService.java
joewhite101: How does it work on the page tree? I'm not talking about the check but the code inside of the block. The calls for set/get seem like they should be elsewhere.

alepintus: Hey @joewhite101 , thanks for the review. 
This code is not duplicated and is present only here (I wrote this from scratch). In this section I'm setting a PagePositionRequest object (to pass it to the movePage method) using the request parameters and calculating the position of the page in the selected branch of the tree. In the "page tree" this object is passed directly from the controller to the movePage method.

engine/src/main/java/org/entando/entando/web/page/validator/PageValidator.java
```diff
@@ -125,7 +132,10 @@ public void validateGroups(String pageCode, PagePositionRequest pageRequest, Err
     public void validatePagesStatus(String pageCode, PagePositionRequest pageRequest, Errors errors) {
         IPage parent = this.getPageManager().getDraftPage(pageRequest.getParentCode()),
                 page = this.getPageManager().getDraftPage(pageCode);
+        logger.debug("page.isOnline() {}", page.isOnline());
```
joewhite101, line 135: If we're going to log these we should include an id or something so we know which page we're logging about.

entando-service:
All is well

PR #886, opened 2020-05-07
ENG-523: update configuration of maven-deploy-plugin (v5.3.0)

eugeniosant:


eugeniosant:
@joewhite101 Do we have to do the same modification on master branch?

joewhite101:
@eugeniosant i've been doing it manually for the formal releases. Snapshots are working without it right now on master. We'd like to get that in everywhere but not critical. There is some refactoring going on around the poms as well.

entando-service:
All is well

PR #893, opened 2020-05-12
ENG-407 Changing how we validate if the current password is correct

BrenoQVDS:
You might noticed that I didn't add tests for this. We currently already have tests for the user changing its password:
	UserControllerIntegrationTest.testUpdatePassword_1() on core
	UserControllerIntegrationTest.testUpdatePassword_2() on core
	UserManagerIT.testUserUpdatePassword() on entando-keycloak-plugin
	
The logic didn't change, just the way we check if an old password is valid or not. This was necessary because we don't have access to what is the current password on keycloak.

After this change the legacy admin console still works, and the app builder works with keycloak and without it.

### DIFF
--- a/engine/src/main/java/org/entando/entando/web/user/validator/UserValidator.java
+++ b/engine/src/main/java/org/entando/entando/web/user/validator/UserValidator.java
@@ -209,8 +209,15 @@ public class UserValidator extends AbstractPaginationValidator {
     }
 
     private boolean verifyPassword(String username, String password) {
-        UserDetails user = this.extractUser(username);
-        return passwordEncoder.matches(password, user.getPassword());
+        try {
+            UserDetails user = this.getUserManager().getUser(username, password);
+            if (user != null) {
+                return true;
+            }
+        } catch (ApsSystemException e) {
+            logger.error("Invalid password for username {}", username, e);
+        }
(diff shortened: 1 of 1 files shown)


entando-service:
All is well

engine/src/main/java/org/entando/entando/web/user/validator/UserValidator.java
```diff
@@ -209,8 +209,15 @@ private void checkNewPassword(String username, String password, BindingResult bi
     }
 
     private boolean verifyPassword(String username, String password) {
-        UserDetails user = this.extractUser(username);
-        return passwordEncoder.matches(password, user.getPassword());
+        try {
+            UserDetails user = this.getUserManager().getUser(username, password);
```
joewhite101, line 213: Does this call throw an exception if the password doesn't match?

BrenoQVDS: Yes. It first sets:
` bindingResult.rejectValue("oldPassword", ERRCODE_OLD_PASSWORD_FORMAT, new String[]{}, "user.password.old.invalid");`
Then:
`throw new ValidationGenericException(bindingResult);`

PR #909, opened 2020-06-17
ENG-66: Definition of default widgets available via PageModel api

eugeniosant:


entando-service:

engine/src/main/java/org/entando/entando/aps/system/services/pagemodel/PageModelService.java
```diff
@@ -1,18 +1,31 @@
+/*
```
Kerruba, line 1: We need to change this to use LGPL 3 instead of LGPL 2.1, 
This is valid for all the other classes too

entando-service:
All is well

engine/src/main/java/org/entando/entando/aps/system/services/pagemodel/PageModelService.java
eugeniosant: Is there a task/ticket where we can do that? I ask that because this modification has to be extended in the whole project and it's not related to the task on this pr

Kerruba: Let's create a ticket for that to make sure all are in sync, but I know the Readme has been all updated to LGPL3 and some classes already use the LGPL3 nomenclature

PR #910, opened 2020-06-23
ENG-731-admin-console-cms- added ellipsis for action dropdown

DMinnai:
ENG-731-admin-console-cms- added ellipsis for action dropdown


entando-service:
All is well

PR #911, opened 2020-06-29
Eng 698 readiness probe endpoint

firegloves:
Added a new endpoint to manage the readiness probe


engine/src/main/java/org/entando/entando/web/health/HealthController.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @GetMapping
+    public ResponseEntity<HealthResponse> isHealthy() {
+
+        logger.debug("health check");
+
+        boolean healthy = this.healthService.isHealthy();
+
+        return new ResponseEntity<>(new HealthResponse(healthy ? STATUS_UP : STATUS_DOWN), HttpStatus.OK);
```
ffleandro, line 50: `STATUS_UP` and `STATUS_DOWN` are never used, since when status is unhealthy an exception is thrown.
Either remove this check and always return `STATUS_UP`, or remove Exception thrown, send false when is not healthy and handle true/false value in here.

Everything else looks good to me!

ffleandro: I prefer the second option, since it is cleaner.
When the status is unhealthy you're handling it as an Error response, but actually nothing wrong has happened. Returning false is a normal response in the flow and responsibility of this endpoint.

firegloves: Ok it makes sense. @Kerruba if you agree too I proceed to apply the requested changes

Kerruba: @firegloves I agree with @ffleandro

entando-service:
All is well

PR #936, opened 2020-10-26
Update maven-dependency-check plugin

pietrangelo:


entando-service:

